PR #34, opened 2018-03-15
Add CancellationToken param to Async methods

StephaneDelcroix:
This adds overloads for all public *Async methods to add a
CancellationToken parameter.
That token is then passed over to the HttpClient.*Async methods,
allowing to cancel a query.

### DIFF
--- a/src/GraphQL.Client/GraphQLClient.cs
+++ b/src/GraphQL.Client/GraphQLClient.cs
@@ -3,6 +3,7 @@ using System.IO;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Text;
+using System.Threading;
 using System.Threading.Tasks;
 using GraphQL.Client.Exceptions;
 using GraphQL.Common.Request;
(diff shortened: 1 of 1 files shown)


deinok:
Nice, can you please do the following in order to use "optional arguments"?
```csharp
cancellationToken = CancellationToken.None
```

StephaneDelcroix:
I chose not to do optional arguments, as it's a breaking change, and all code compiled against a previous version of the graphql-client will stop working, unless it's recompiled.

given this information, tell me if you still want me to do optional args, or not.

deinok:
Yes, the default arguments should be okey in order to avoid having more methods.
But unless using reflection that should not be a breaking change, i think.

StephaneDelcroix:
the current signature of `GetQueryAsync` is

```
.method public hidebysig 
        instance class [mscorlib]System.Threading.Tasks.Task`1<class GraphQL.Client.GraphQLResponse> GetQueryAsync (
            string query
        ) cil managed 
```

and all existing compiled call invokes it as such
```
...
ldstr "foo"
call instance class [mscorlib]System.Threading.Tasks.Task`1<class GraphQL.Client.GraphQLResponse> GraphQL.Client.GraphQLClient::GetQueryAsync(string)
```

if we add an optional parameter `cancellationToken = default(CancellationToken)` (and not `cancellationToken = CancellationToken.None` as `None` is not a `const` ), the signature becomes
```
.method public hidebysig 
        instance class [mscorlib]System.Threading.Tasks.Task`1<class GraphQL.Client.GraphQLResponse> GetQueryAsync (
            string query,
            [opt] valuetype [mscorlib]System.Threading.CancellationToken cancellationToken
        ) cil managed 
```

and all code needs to be recompiled in order to changed the IL to
```
...
ldstr "foo"
ldloca.s 1
initobj [mscorlib]System.Threading.CancellationToken
ldloc.1
call instance class [mscorlib]System.Threading.Tasks.Task`1<class GraphQL.Client.GraphQLResponse> GraphQL.Client.GraphQLClient::GetQueryAsync(string, valuetype [mscorlib]System.Threading.CancellationToken)
```

This shows that optional parameters are only compiler-time syntactic sugar, but the runtime doesn't care much about it.

In which case could this pose a problem ?
Let's say someone creates a lib that depends on graphQL.Client, and is distributed as binary only. a second developer, using that lib, might at some point update the graphql-client (without a new version of the lib), and that lib will stop working.

This is why adding optional parameters is, in the same way as changing a method signature, a breaking change, when adding an overload is not (adding an overload could break reflection code using `Single` to get the method by name only, and not signature, but  you shouldn't care about it ¯\_(ツ)_/¯ )

deinok:
I'm not an expert in IL code but LGTM.

Give me a little bit more of time to make some checkings

PR #48, opened 2018-05-30
HTTP Status Code in the GraphQLResponse.

mtoporek:


deinok:
ref: https://github.com/graphql-dotnet/graphql-client/blob/master/src/GraphQL.Client/GraphQLClient.cs#L184

In case of an invalid GraphQL Response two things can happen.
A JSON exception when code is 2XX
A GraphQLException with a HttpMessageResponse inside.

So, your Unauthorized case should trigger the GraphQLException.
Is it not thrown?

mtoporek:
No it is not thrown. I have checked it also with your unit tests and it goes to the return section.

Issue #53, opened 2018-08-07
Make GetAsync preserve Endpoint query params such as API keys


mjul:
# Problem
`GetAsync` discards query parameters specified in the service URI  `GraphQLHttpClientOptions.Endpoint`. 

# Motivation
This is a problem if we are working with _e.g._ service endpoints that require an API key. Azure Functions are one such example. Here the API key can be provided through query parameter in the URI or an HTTP header. An example of such an endpoint  URI with key is `https://graphqlapi-dev.azurewebsites.net/api/GraphQLService?code=Xp12xdeP3SrF8H/CSaWk5m9Xyf0E5S4gyEF/ex/2L1wEbV5oAPdLQg==` 

If we send a query to that endpoint `GetAsync` will send a query to  `https://graphqlapi-dev.azurewebsites.net/api/GraphQLService?query=...` without the API key in the `code` parameter.
This means access will be denied, making the request fail.

The expected URI would be `https://graphqlapi-dev.azurewebsites.net/api/GraphQLService?code=Xp12xdeP3SrF8H/CSaWk5m9Xyf0E5S4gyEF/ex/2L1wEbV5oAPdLQg==&query=...`

# Cause
The problem is found in the `GetAsync` method on `GraphQLHttpHandler`:

```csharp
		public async Task<GraphQLResponse> GetAsync(GraphQLRequest request, CancellationToken cancellationToken = default) {
			if (request == null) { throw new ArgumentNullException(nameof(request)); }
			if (request.Query == null) { throw new ArgumentNullException(nameof(request.Query)); }

			// THE PROBLEM IS THE FOLLOWING CODE  
			var queryParamsBuilder = new StringBuilder($"query={request.Query}", 3);
			if (request.OperationName != null) { queryParamsBuilder.Append($"&operationName={request.OperationName}"); }
			if (request.Variables != null) { queryParamsBuilder.Append($"&variables={JsonConvert.SerializeObject(request.Variables)}"); }
			using (var httpResponseMessage = await this.HttpClient.GetAsync($"{this.Options.EndPoint}?{queryParamsBuilder.ToString()}", cancellationToken).ConfigureAwait(false)) {
				return await this.ReadHttpResponseMessageAsync(httpResponseMessage).ConfigureAwait(false);
			}
		}
```

# Context
For comparison, it looks like `PostAsync` already preserves the query parameters in the URI since it uses the externally provided service URI in `GraphQLHttpClientOptions.Endpoint` without modification.

[...]

PR #55, opened 2018-08-16
Add tests preserving query params in endpoint (See #53)

mjul:
This adds a single unit test for the preservation of endpoint query parameters in the new client


deinok:
@mjul Does this mean that your issue is solved in the source code of master?

mjul:
It is only solved for the new client.

For the old client (now obsolete) on which the bug was reported, the bug is still there and it is not directly testable as its ...Options class no longer has a member for the HttpMessageHandler so I can pass in a stub that would give me access to the outgoing request to verify the URI parameters.

I could add it if you like, or add some kind of override for test purposes if you prefer that.

Cheers
Martin

PR #57, opened 2018-08-24
Cleanup

SimonCropp:


SimonCropp:
whenever you `await` the compiler generates a state machine to handle the underlying complexity. u can see this if u use a low level decompiler. there is an overhead to this state machine https://www.red-gate.com/simple-talk/dotnet/net-framework/the-overhead-of-asyncawait-in-net-4-5/

if you have only one async action, and u have no dispose (that needs to happen after the async completes), then you can do a direct return of the Task. hence avoiding the  `await` and the resultant state machine. this results in 

 * faster execution at runtime
 * simpler stack traces hence easier to debug
 * faster compilation (although prob difficult to notice)
 * smaller resultant assembly
 * faster Jit
 * less code

deinok:
@SimonCropp I have double checked it and seems like you are right, the MSIL code generated seems more performant. Thanks for the advice

PR #58, opened 2018-08-25
Remove forced camel casing

sys-dev:
Forcing camel casing for all elements by using CamelCasePropertyNamesContractResolver() precludes use of PascalCasing for names of data fields. By changing the names of query, operationName and variables to to comply with the spec the CamelCasePropertyNamesContractResolver is not required.

### DIFF
--- a/src/GraphQL.Client/Http/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/Http/GraphQLHttpClient.cs
@@ -100,26 +100,26 @@ namespace GraphQL.Client.Http {
 			this.graphQLHttpHandler = new GraphQLHttpHandler(options,httpClient);
 		}
 
-		public async Task<GraphQLResponse> SendQueryAsync(string query, CancellationToken cancellationToken = default) =>
-			await this.SendQueryAsync(new GraphQLRequest { Query = query }, cancellationToken).ConfigureAwait(false);
+		public async Task<GraphQLResponse> SendQueryAsync(string _query, CancellationToken cancellationToken = default) =>
+			await this.SendQueryAsync(new GraphQLRequest { query = _query }, cancellationToken).ConfigureAwait(false);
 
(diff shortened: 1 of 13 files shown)


deinok:
Hi, C# have some naming conventions. This is the reason that `CamelCasePropertyNamesContractResolver` exists. To have the C# naming conventions, and follow the spec of GraphQL.

With the changes that you have requested, the lib will follow the spec of GraphQL but not the C# naming conventions.

Can you give a better reason to accept this change?

sys-dev:
So if the intention is for graphql-dotnet/graphql-client library to work 
with widely used GraphQL server-side implementations, and allow full 
range of data field naming, then it seems to violate C# naming 
convention in this small area is best trade-off.

There is discussion here regarding this issue:
https://github.com/facebook/graphql/issues/209

 >> However any field or type not described directly in the GraphQL spec 
can be cased however you like.

Without the proposed changes, the CamelCasePropertyNamesContractResolver 
is required to force Query to query, OperationName to operationName and 
Variables to variables in the outbound JSON generated by JSON.Net.  
Without proper casing of these, the server rejects the request.  
However, using CamelCasePropertyNamesContractResolver on the entire 
payload, also changes *data field names* to camelCase.  Data fields (per 
spec) can be any desired casing, so the current graphql-client library 
won't work where data field names are not camel case.

For reference, the Apollo Server 2 server docs show JSON for a valid POST:
https://www.apollographql.com/docs/apollo-server/requests.html

deinok:
> However, using CamelCasePropertyNamesContractResolver on the entire
payload, also changes *data field names* to camelCase.  Data fields (per
spec) can be any desired casing, so the current graphql-client library
won't work where data field names are not camel case.

Okey, you are right, this is an issue.
So for me the solution is force the `Query`, `OperationName` and `Variables` to CamelCase and disable the `CamelCasePropertyNamesContractResolver` by default.

So, I will open an issue to handle this in a way that is correct in C# and with the spec of GraphQL

ref: http://facebook.github.io/graphql/June2018/#sec-Names

Issue #25, opened 2018-02-13
[Subscriptions] - Design


deinok:
Enable the feature of sending subscriptions like query.

Open discusion about what is the best way to implement it, API, and technologies.

deinok:
If using WebSockets there are two viable ways:
* https://github.com/sta/websocket-sharp & https://www.nuget.org/packages/WebSocketSharp
* System.Net.WebSockets.ClientWebSocket

saba-sabrin:
Hi, I have tried the Subscription example to subscribe for a GraphQL subscription request. But, apparently, it is not receiving any notifications for the subscription. Could you please recheck the example and provide a concrete solution on how we can subscribe properly and listen to the incoming notifications? Would be really great and appreciated !! Looking forward to your reply. Thanks.

deinok:
@saba-sabrin Todey i will create a sample

saba-sabrin:
@deinok This is my code for using the subscription based on provided subscription usage:
ref:  graphql-client/samples/GraphQL.Client.Sample/Program.cs

I can't share the GraphQL API details as of it belongs to my client and I'm using their private server. The API also uses token based authentication. That's why I need an example of proper usage of subscription and notification reception.

This is my GraphQL query:

var Query = @"subscription {
                        deviceCharacteristic {
                            node {
                            characteristicType
                            characteristicValue
                            deviceId
                            deviceDisplayName
                        }
                    }
                }";

            // Simple delegate for checking "OnReceive" event data
            EventCheck eventCheckHandler = EventChecker;

// Function for subscription request using GraphQL Client
using (var graphQLHttpClient = new GraphQLHttpClient("ws://...."))  // Websocket based URL
            {
                var subscriptionResult = graphQLHttpClient.SendSubscribeAsync(Query).Result;
                subscriptionResult.OnReceive += (res) => {
                    Console.WriteLine("received event.");
                    eventCheckHandler(res);
                };
                
                Console.ReadKey();
            }
    
 // Delegate declaration
public delegate void EventCheck(GraphQLResponse res);

        public void EventChecker(GraphQLResponse res)
        {
            var response = res.Data;
            Console.WriteLine("Event detected");
        }

deinok:
Try this and report back.
you set the event handler but the background thread that reads the websocket is not started
```csharp
            var subscriptionResult = graphQLHttpClient.SendSubscribeAsync(Query).Result;
            subscriptionResult.StartAsync();
[...]

PR #80, opened 2018-11-07
Observable Subscriptions

rose-a:
Relating to #25

@deinok Here is my first stab at creating a subscription "API" based on `System.Reactive`.

I've left the existing API untouched and implemented my idea alongside it. I've also added a simple integration test based on the chat sample to prove it working.

Currently the new variant could be used as follows:

```c#
var graphQLRequest = new GraphQLRequest
{
    Query = @"
    subscription {
        messageAdded{
        content
        }
    }"
};

// create the desired subscription stream. 
// No connection is established yet.
IObservable<GraphQLResponse> subscriptionStream = client.CreateSubscriptionStream(graphQLRequest);

// subscribe to the stream. When the first subscription is made, 
// the web socket connection for the stream is initiated
IDisposable subscription = subscriptionStream.Subscribe(response => processUpdate(response));

// when you are done with the subscription, dispose of it. 
// Disposing the last subscription will dispose the web socket and all related resources.
subscription.Dispose();
```

Disposing the parent `GraphQlHttpClient` will cancel and dispose all web sockets made on it, which will result in the subscribers receiving an `OnError()` call with a `ThreadCancelledException`.  

This should satisfy the requirements you mentioned in #25.

Please let me know what you think.


deinok:
Looks pretty good.

deinok:
And about the integration tests, probably we are going to move to use our own integrated server and not something exposed to internet

rose-a:
The integration tests are already using the "IntegrationTestServer" in the tests subdirectory (since I did not find any public demo APIs which support subscriptions).

So basically this server project can also be reused for all other tests which currently run against a public API.

For everything NOT using websockets we can do this the way pointed out in [this Microsoft tutorial](https://docs.microsoft.com/de-de/aspnet/core/test/integration-tests?view=aspnetcore-2.1#basic-tests-with-the-default-webapplicationfactory)

With `ClientWebSocket` the `WebApplicationFactory` doesn't seem to work, so we have to create the test server the "manual" way.

deinok:
@rose-a Very well done.
It need a little bit more reviewing but Reactive Extensions seems the way to go.
I temporaly move it to `subscriptions-api` while we ensure that API is "perfect".

Many thanks. ( Paypal and I will give you the beer I promised ;) )

curbol:
Can you create a new alpha with this included? I would like to use it.

akoliya003:
> Relating to #25
> 
> @deinok Here is my first stab at creating a subscription "API" based on `System.Reactive`.
> 
> I've left the existing API untouched and implemented my idea alongside it. I've also added a simple integration test based on the chat sample to prove it working.
> 
> Currently the new variant could be used as follows:
> 
> ```cs
> var graphQLRequest = new GraphQLRequest
> {
>     Query = @"
>     subscription {
>         messageAdded{
>         content
>         }
>     }"
> };
> 
> // create the desired subscription stream. 
> // No connection is established yet.
> IObservable<GraphQLResponse> subscriptionStream = client.CreateSubscriptionStream(graphQLRequest);
> 
> // subscribe to the stream. When the first subscription is made, 
> // the web socket connection for the stream is initiated
> IDisposable subscription = subscriptionStream.Subscribe(response => processUpdate(response));
> 
> // when you are done with the subscription, dispose of it. 
> // Disposing the last subscription will dispose the web socket and all related resources.
> subscription.Dispose();
> ```
> 
> Disposing the parent `GraphQlHttpClient` will cancel and dispose all web sockets made on it, which will result in the subscribers receiving an `OnError()` call with a `ThreadCancelledException`.
> 
> This should satisfy the requirements you mentioned in #25.
> 
> Please let me know what you think.

Hi, it seems like CreateSubscriptionStream is not a member function of GraphQL.Client.Http.GraphQLHttpClient

I am using GraphQL (3.0.0-preview-1271) , GraphQL.Client(2.0.0-alpha.3) , GraphQL.Common (2.0.0-alpha.3).

Could you please send me sample link.

Appreciate your help. 

Thanks,

PR #84, opened 2018-12-10
Use single websocket for all subscriptions

rose-a:
I've worked a bit on the API and restructured it to use a single web socket for all subscriptions made on a `GraphQLHttpClient`.

I've also added some ways to customize error handling and to automatically reconnect in case of a connection loss without having to re-create the whole subscription.

From here we're also not far from being able to use the client in "full websocket mode", meaning that regular queries and mutations also would use the websocket connection instead of sending separate POST requests for each query. This is also supported by the Apollo GraphQL server and the [graphql-dotnet/server](https://github.com/graphql-dotnet/server) middleware.

What's still missing is an automated test to confirm the reconnect working, because I did not find any way to programmatically restart Kestrel within a XUnit test. I had to test this one manually using the `IntegrationTestServer` app and the `SubscriptionIntegrationTest.ConsoleClient` console app.

This PR relates to #25


deinok:
Hi, @rose-a I'm porting all the code to use https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/nullable-reference-types.
You will need dotnet-sdk-3.0-preview1 atleast, for the IDE VS2019-preview should work well. You dont need VS2019 but other editors could fail at detecting the nullable-references

PR #91, opened 2019-01-23
Fix StackOverflowException in GraphQLRequest.GetHashCode()

rose-a:
The original `GetHashCode()` method resulted in a `StackOverflowException` on usage.

I created a new implementation (with a little assistance from ReSharper) which seems to work so far.


deinok:
Hey, good catch, can you change the origin to dev?

PR #92, opened 2019-01-28
Subscriptions API

rose-a:
### Single websocket per client

There is only a single websocket connection established per client. This connection is shared across all subscriptions and optionally requests.

### Requests via websocket

By setting `GraphQLHttpClientOptions.UseWebSocketForQueriesAndMutations = true` you can force requests and mutations to use the websocket too, which can reduce the connection overhead on regular requests.

### Re-establish subscriptions on connection loss

If the receiving pipeline of the WebSocket encounters a *WebSocketException*, the client tries to re-establish the connection and on success re-initiates all current subscriptions to the server. 

Since this prevents all *WebsocketExceptions* from reaching the Observer, all exceptions occurring in the receiving pipeline are submitted to the `GraphQLHttpClient.WebSocketReceiveErrors` observable (i.e. for logging purposes.

This behaviour can be overridden per subscription by using the `GraphQLHttpClient.CreateSubscriptionStream(GraphQLRequest request, Action<Exception> exceptionHandler)` overload and re-throwing the exception inside the *exceptionHandler*

### Reuse subscriptions for identical requests
The created observables for subscriptions are cached in `GraphQLHttpClient`. If a second subscription is made with the identical request, the existing observable is returned (instead of creating a second subscription to the server)

### Receive long response messages from the WebSocket

The receiving loop now keeps reading into a smaller buffer until `WebSocketReceiveResult.EndOfMessage` becomes `true`.

### Windows 7 compatibility

The usage of `ClientWebSocket` throws a `PlatformNotSupportedException` on Windows 7 (since Microsoft did not implement it there).

This is fixed by using https://github.com/PingmanTools/System.Net.WebSockets.Client.Managed


halitizgin:
Thanks. This package does not appear in NuGet. How can we use this package? I'm sorry I don't have much information about the this package.

PR #93, opened 2019-01-31
Added QueryBuilder

gregsdennis:
The current way of supplying a query/mutation to a request is with strings.  This is prone to errors as it is not checked by the compiler.

This change adds a query builder modeled after Entity Framework Core's navigation properties.

I've added support for generating strings for:

- Queries
- Mutations
- Query/Mutation Arguments
- Field Arguments

I intend on expanding on this (fragments, etc.), but what's here is a good start.

### DIFF
--- a/src/GraphQL.Common/GraphQL.Common.csproj
+++ b/src/GraphQL.Common/GraphQL.Common.csproj
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Project Sdk="Microsoft.NET.Sdk">
 
 	<Import Project="../src.props" />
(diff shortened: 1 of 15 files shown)


deinok:
Really good job, I will need time to review its API, but looks pretty good

akash-geon:
@gregsdennis  Do we have an ETA on when this feature would be released?

gregsdennis:
I don't know. I was just offering it as a solution. There was a lot of discussion around ideas to solve the problem.

I'm not actually part of this project.

src/GraphQL.Common/GraphQL.Common.csproj
```diff
@@ -12,6 +12,7 @@
 	</PropertyGroup>
 
 	<ItemGroup>
+		<PackageReference Include="Humanizer" Version="2.5.16" />
```
SimonCropp: can the extra dependency be ignored?

gregsdennis: It performs some conversions between the code property names and the schema field names, primarily things like camel-casing. You'd have to invite an implementation of that locally.

gregsdennis:
I'm going to close this in favor of #111.  I rather like that better.

PR #103, opened 2019-03-28
Expose the HttpClient on GraphQLHttpClient

rose-a:
Hi, I went on and exposed the `HttpClient` from inside the `GraphQLClient` and also made the constructor accepting a `HttpClient` public (its "exposed" by the `HttpClientExtensions` anyway).

This makes it easier to use the library in combination with DI systems.

Exposing the `HttpClient` as a readonly property has the advantage that existing extension methods from other libraries (like from the [IdentityModel library](https://github.com/IdentityModel/IdentityModel2/blob/master/src/Client/HttpClientExtensions.cs)) can be used with `GraphQLHttpClient` as well.

### DIFF
--- a/src/GraphQL.Client/Http/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/Http/GraphQLHttpClient.cs
@@ -19,7 +19,12 @@ namespace GraphQL.Client.Http {
 		/// <summary>
 		/// Gets the headers which should be sent with each request.
 		/// </summary>
-		public HttpRequestHeaders DefaultRequestHeaders => this.graphQLHttpHandler.HttpClient.DefaultRequestHeaders;
+		public HttpRequestHeaders DefaultRequestHeaders => this.HttpClient.DefaultRequestHeaders;
+
+		/// <summary>
+		/// the instance of <see cref="HttpClient"/> which is used internally
+		/// </summary>
+		public HttpClient HttpClient => this.graphQLHttpHandler.HttpClient;
 
 		/// <summary>
 		/// The GraphQL EndPoint to be used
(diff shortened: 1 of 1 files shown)


PeterTognola:
Hi, this would also be great to have. Is there any plan on completing this pull request?

raphaabreu:
Hello @deinok, how can I help you to get this PR approved and available on a nuget package? I am using this client in a scenario with 10k requests/s and the fact that I cannot reuse an HttpClient from the IHttpClientFactory is totally killing my performance. For now I have no choice but to send the http request myself but I would really want to do this through the GraphQL Client.

Just give me the directions and I'll make the time to get it done.

wc-matteo:
I also need this. Can we make it happen? @deinok

raphaabreu:
I ended up forking and making available a version that is a simple wrapper around an HttpClient. We had several issues when we started to load test the usage of this client because of the way the original package created the http message handler inside the options object, with no way of reusing or properly disposing it, causing socket exhaustion.

The wrapper version I made is exactly the same except for the fact that you have to provide a fully configured httpclient to it.

https://www.nuget.org/packages/GraphQL.HttpClientWrapper/

wc-matteo:
Sounds awesome, thanks @raphaabreu !

benmccallum:
We're getting socket exhaustion too, and this might be the culprit. Thanks so much for writing this and releasing via NuGet. Hopefully this project, under the `graphql-dotnet` banner, gets fixed; as that's really not the way to handle HttpClient, for x-compat (not just targeting corefx) it should really be requiring a HttpClient be passed in so the consumer can manage its lifetime and in corefx we can register it as a typed client.

rose-a:
incorporated in #152

PR #104, opened 2019-04-01
Configurable preprocessing method for requests

rose-a:
This PR adds a configurable preprocessing method to `GraphQLHttpOptions`, which is executed for each request prior to sending it. By default it just passes on the received request, so the default behaviour is not changed.

The main intention for me is to automatically add the authorization info which was set in the `DefaultRequestHeaders` to a request payload which is sent via websocket (so no HTTP headers are available).

There might be a lot of other usecases though....


rose-a:
incorporated in #152

PR #109, opened 2019-04-29
Update readme

omerlh:
The example in the readme does not compile - look like the query is now a property, and not a constructor property


deinok:
@omerlh In v2 aka master this has changed

PR #112, opened 2019-05-05
Feature: expose response headers

### LINKED ISSUE #90: Retrieving Headers from Response
I have been having a problem where i have been trying to get headers from GraphQLResponse. And i have seen that it is possible for example in [HttpRequest#Headers](https://docs.microsoft.com/en-us/dotnet/api/system.web.httprequest.headers?view=netframework-4.7.2), but not in here, like for example i have been trying to get headers from [Anilist](https://anilist.gitbook.io/anilist-apiv2-docs/overview/rate-limiting) so i could stop myself getting rate limited.

Is there a way to get headers from `GraphQLResponse` or `GraphQLHttpClient`?

willwolfram18:
Included a new `Headers` property on the `GraphQLResponse` that allows users access to the response headers for their request to the server. The property uses the `IEnumerable<KeyValuePair<string, IEnumerable<string>>>` because the `HttpResponseHeaders` class implements that interface, and prevents the `GraphQLResponse` from getting directly tied to the `System.Net.Http` library.

The implemented unit test is designed to simply ensure that the Headers property is not null following the HTTP request, since the `HttpResponseMessage.Headers` property is a get-only property, and cannot be mocked by something like Moq because it is not virtual.

This PR closes #90


sungam3r:
Headers are not a part of GraphQL response by spec. It refers to HTTP transport implementation details and should be handled respectively.

rose-a:
response headers are exposed since V3.0.0

Issue #139, opened 2019-11-07
Request to add API that returns string responses/errors rather than deserialized results


sidhants:
This will help in making the library usable with il2cpp/no dynamics/WebGL based use cases including Unity which is a big community. The consumer of the API can perform the deserialized operation on their side as needed with any compatible package. 

A small addition to the source code at my end was enough to get the responses to work nicely-
`public Task<string> SendQueryAsync(GraphQLRequest request, CancellationToken cancellationToken = default) =>
			this.graphQLHttpHandler.PostAsync(request, test, cancellationToken);`

il2cpp successful results in Unity-
![image](https://user-images.githubusercontent.com/18451476/68371919-99c3fd00-00f4-11ea-9ad9-6c67fe3f73bc.png)

@deinok Please let me know what you think about this? I feel it can be very useful to increase the scope of use. Serialization does not have any issues currently with il2cpp once newton json is configured to turn off dynamics and works out of the box as expected. But deserialization does not work because of how the graphql response class is set up and dependent on dynamics data.
The version I am using is 'GraphQL.Client.2.0.0-alpha.3'.

rose-a:
@sidhants please check the current API (`develop` branch), it uses *System.Text.Json* instead of *Json.Net* and does not use dynamics anymore.

sidhants:
> @sidhants please check the current API (`develop` branch), it uses _System.Text.Json_ instead of _Json.Net_ and does not use dynamics anymore.

@rose-a I am going to give it another shot with Unity integration this weekend. Though I have heard System.Text.Json is not compatible with Unity probably due to reflection features not supported on il2cpp/iOS/VR/AR headsets. I will confirm this, otherwise need to think of another solution.

rose-a:
@sidhants the biggest problem i discovered are the `Extensions` fields on `GraphQLResponse` and `GraphQLError`.

I'm not keen on having to specify 3 types for every request...

rose-a:
Changed the type of extensions to `System.Text.Json.JsonElement` in e1c8f3b1ac5ffbfcec9dbf6d36f1aad6bcacf65b. 

If the issue is just the dynamics this should solve it.

DamianMehers:
System.Text.Json is also not currently supported by Xamarin iOS (or WatchOS) so returning raw JSON would help there too https://github.com/dotnet/runtime/issues/31326

rose-a:
Maybe a dumb question, but how does the serialization work if the serializer is not supported?

[...]

Issue #59, opened 2018-08-27
CamelCasePropertyNamesContractResolver causes incompatibility with Spec


deinok:
ref: http://facebook.github.io/graphql/June2018/#sec-Names
ref: https://github.com/graphql-dotnet/graphql-client/pull/58

> However, using CamelCasePropertyNamesContractResolver on the entire
payload, also changes data field names to camelCase. Data fields (per
spec) can be any desired casing, so the current graphql-client library
won't work where data field names are not camel case.

So, the idea is force `Query`, `OperationName` and `Variables` to CamelCase without forcing the usage `CamelCasePropertyNamesContractResolver` that causes that the model of Variables to be forced to be CamelCase.

This needs further investigation, like:
* Requests: It causes renaming of variables?
* Response: Is the response also afected by this?
* Serialization: Is there a way to fix this issues without depending on JSON implementation? Other Serialization formats cases

CC: @sys-dev

PR #160, opened 2020-02-11
Pluggable JSON Serializer

### LINKED ISSUE #137: Dynamic types not supported in Unity with IL2CPP scripting backend causing errors.
I have been trying to use the client with Unity 3D but I am having issues since Unity does not support dynamic types (IL2CPP scripting backend) and thus errors out. Is there any way I can still use the client without using dynamics?I am restricted by certain platforms like iOS and Magic Leap which only work with IL2CPP currently.
That being said, I have had success using the client on Windows and android devices through unity and everything works well.
Would love to explore any options to still get this library to work.
Some of the other errors that I see with the IL2CPP:

2019-10-23 20:29:53.207 21466-21480/? I/Unity: try send req
    UnityEngine.Logger:Log(LogType, Object)
    Main:.ctor()
    System.Net.Http.Headers.HttpHeaders:SetValue(String, T, Func`2)
    Main:TestRequest()
     
    (Filename: ./Runtime/Export/Debug/Debug.bindings.h Line: 48)
2019-10-23 20:29:53.257 21466-21480/? I/Unity: exceptionError getting value from 'Query' on 'GraphQL.Common.Request.GraphQLRequest'.
    UnityEngine.Logger:Log(LogType, Object)
    Main:.ctor()
    System.Net.Http.Headers.HttpHeaders:SetValue(String, T, Func`2)
    Main:TestRequest()

rose-a:
This moves the JSON (de-)serialization out of *GraphQL.Client* to allow replacing the underlying JSON library.

Closes #137, might render #139 obsolete.


src/GraphQL.Client.Abstractions.Websocket/GraphQL.Client.Abstractions.Websocket.csproj
```diff
@@ -0,0 +1,12 @@
+<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
```
sungam3r: Why do you use netstandard2.1? What APIs from it do you use?

sungam3r:
Relates to graphql-dotnet/server#299 , @rose-a you can look at different optimizations made in server project while moving serializers out of main code.

src/GraphQL.Client.Abstractions.Websocket/GraphQL.Client.Abstractions.Websocket.csproj
rose-a: I guess that's a leftover from me fighting with Visual Studio... Will be removed....

rose-a:
Thanks for the hint... odd that we're doing the same work there on 2 sides...

benmccallum:
> Thanks for the hint... odd that we're doing the same work there on 2 sides...

Three sides if you include the [graphql-dotnet IDocumentWriter PR](https://github.com/graphql-dotnet/graphql-dotnet/pull/1512) haha.

src/GraphQL.Client.Abstractions.Websocket/GraphQL.Client.Abstractions.Websocket.csproj
```diff
@@ -0,0 +1,13 @@
+<Project Sdk="Microsoft.NET.Sdk">
+	<Import Project="../src.props" />
+
+	<PropertyGroup>
+	<TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
```
sungam3r: netstandard2.1

rose-a:
The *System.Text.Json* variant currently only works with `JsonNamingPolicy.CamelCase` (which can cause a spec incompatibility, see #59). 

For *Newtonsoft.Json* this is solved by adding the `[DataMember(Name = "name")]` attributes (which are independent of the JSON library.

I saw the solution chosen for `GraphQLRequest` in graphql-dotnet/server#299, but I'm not quite sure how this could be implemented for `GraphQLResponse` and `GraphQLError` without adding yet another generic to GraphQLResponse (to be able to replace the type of `GraphQLError` inside)...

Ideas welcome! 😉

sungam3r:
> The System.Text.Json variant currently only works with JsonNamingPolicy.CamelCase

What does that mean? You can use any policy you want.

rose-a:
If you don't use `JsonNamingPolicy.CamelCase`, then `Data`, `Variables`, `OperationName` will be serialized as they are (with starting capital), which is not conform with the GraphQL spec.

If you use it, then you cannot send variables containing capital fields (which would be ok with the spec)

sungam3r:
1. You can use different policies for serialization and deserialization.
2. `JsonSerializerOptions.DictionaryKeyPolicy` or `JsonSerializerOptions.PropertyNameCaseInsensitive` may help.

rose-a:
@sungam3r Anything that should be changed in your opinion?

sungam3r:
Firstly, I am glad that work on this project has continued. I see a lot of changes taking place. I remember very well that until recently there were a lot of not implemented features and bugs to fix. My attitude to the current process is this: I don’t really go into the details of the changes and wait for the project to stabilize. After that, I have plans to make a code review. In my projects, I completely refused to use this package just because of the lack of support. I have my own implementation that suits me. Nevertheless, if I see that the package really matured, then most likely I want to use it again and throw out my own code. That is why I intend to make a review in the future. In the meantime, I avoid watching PRs with dozens of modified files. Honestly, when I see the number 83 on the modified files tab header, I’m even scared to click on the link.

rose-a:
Ok, no worries :wink:

PR #164, opened 2020-02-18
Websocket connection status

rose-a:
Provides an observable status for the websocket inside `GraphQLHttpClient` as well as a `OnWebsocketConnected` callback which is awaited inside `GraphQLHttpWebSocket` after successfully establishing a connection and before sending any requests.


src/GraphQL.Client/GraphQLHttpClientOptions.cs
```diff
@@ -49,5 +49,10 @@ public class GraphQLHttpClientOptions {
 		/// Request preprocessing function. Can be used i.e. to inject authorization info into a GraphQL request payload.
 		/// </summary>
 		public Func<GraphQLRequest, GraphQLHttpClient, Task<GraphQLRequest>> PreprocessRequest { get; set; } = (request, client) => Task.FromResult(request);
+
+		/// <summary>
+		/// This function is called after successfully establishing a websocket connection but before any regular request is made.
```
sungam3r:
```suggestion
		/// This callback is called after successfully establishing a websocket connection but before any regular request is made. 
```

PR #191, opened 2020-03-15
Allow access to HttpResponseHeaders

### LINKED ISSUE #90: Retrieving Headers from Response

rose-a:
fixes #90

### DIFF
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -66,11 +66,10 @@ namespace GraphQL.Client.Http {
 
 		/// <inheritdoc />
 		public async Task<GraphQLResponse<TResponse>> SendQueryAsync<TResponse>(GraphQLRequest request, CancellationToken cancellationToken = default) {
-			if(Options.UseWebSocketForQueriesAndMutations)
+			if (Options.UseWebSocketForQueriesAndMutations)
 				return await this.graphQlHttpWebSocket.SendRequest<TResponse>(request, cancellationToken);
-			
-			var response = await this.SendHttpPostRequestAsync<TResponse>(request, cancellationToken);
-			return response.Response;
+
+			return await this.SendHttpPostRequestAsync<TResponse>(request, cancellationToken);
 		}
 
 		/// <inheritdoc />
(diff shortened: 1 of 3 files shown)


src/GraphQL.Client/GraphQLHttpClient.cs
```diff
@@ -64,10 +65,12 @@ public class GraphQLHttpClient : IGraphQLClient {
 		#region IGraphQLClient
 
 		/// <inheritdoc />
-		public Task<GraphQLResponse<TResponse>> SendQueryAsync<TResponse>(GraphQLRequest request, CancellationToken cancellationToken = default) {
-			return Options.UseWebSocketForQueriesAndMutations
-				? this.graphQlHttpWebSocket.SendRequest<TResponse>(request, cancellationToken)
-				: this.SendHttpPostRequestAsync<TResponse>(request, cancellationToken);
+		public async Task<GraphQLResponse<TResponse>> SendQueryAsync<TResponse>(GraphQLRequest request, CancellationToken cancellationToken = default) {
+			if(Options.UseWebSocketForQueriesAndMutations)
```
sungam3r:
```suggestion
			if (Options.UseWebSocketForQueriesAndMutations)
```

rose-a:
I changed it to `GraphQLHttpResponse<T> : GraphQLResponse<T>`, IMO this is a really clean solution.

There is an extension method `AsGraphQLHttpResponse()` to cast the returned `GraphQLResponse<T>` back to a `GraphQLHttpResponse<T>` (necessary if a dynamic type was used) which will throw when the cast is not possible (aka using a websocket).

sungam3r:
Indeed, more clean.

PR #197, opened 2020-03-20
Fix numeric deserializetion in extensions

### LINKED ISSUE #194: GraphQLExtensionsConverter should handle long/ulong/etc. numeric types as well as int.
see https://github.com/graphql-dotnet/graphql-client/issues/193#issue-584196659

rose-a:
fixes #194 

`GraphQLExtensionsType` renamed to `Map` because that's what it means to represent...


src/GraphQL.Client.Serializer.Newtonsoft/MapConverter.cs
```diff
@@ -69,6 +61,8 @@ private IEnumerable<object> ReadArray(JToken element)
             }
         }
 
+        private object ReadNumber(JToken token) => ((JValue) token).Value;
```
sungam3r: I do not see this method being called.

rose-a: Taken care of in 19382ad

PR #198, opened 2020-03-20
Remove default serializer

### LINKED ISSUE #188: Remove default JsonSerializer and require explicit configuration
This line always overwrites Option.JsonSerializer:

https://github.com/graphql-dotnet/graphql-client/blob/a72700006a0d5fb797bdc0d1d2bc0708556ebb09/src/GraphQL.Client/GraphQLHttpClient.cs#L57

Therefore the following code does not work as expected:

```
_client = new GraphQLHttpClient(o =>
{
    o.EndPoint = new Uri(_testUrl);
    o.JsonSerializer = new SystemTextJsonSerializer(
        options =>
        {
            options.Converters.Add(new JsonStringEnumConverter());
        });
});
```

The constructor above calls the constructor

https://github.com/graphql-dotnet/graphql-client/blob/a72700006a0d5fb797bdc0d1d2bc0708556ebb09/src/GraphQL.Client/GraphQLHttpClient.cs#L53

which sets the JsonSerializer always to `NewtonsoftJsonSerializer`.

By the way, I think it would be better to make a breaking change, remove the reference to `GraphQL.Client.Serializer.Newtonsoft` and require the caller to provide the serializer.

rose-a:
fixes #188


tests/GraphQL.Integration.Tests/GraphQL.Integration.Tests.csproj
```diff
@@ -28,4 +28,8 @@
     <ProjectReference Include="..\IntegrationTestServer\IntegrationTestServer.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <Folder Include="Extensions\" />
+  </ItemGroup>
```
sungam3r:
```suggestion
```

tests/GraphQL.Integration.Tests/Helpers/WebHostHelpers.cs
```diff
@@ -41,9 +41,9 @@ public static GraphQLHttpClient GetGraphQLClient(int port, string endpoint, bool
             => new GraphQLHttpClient(new GraphQLHttpClientOptions
             {
                 EndPoint = new Uri($"http://localhost:{port}{endpoint}"),
-                UseWebSocketForQueriesAndMutations = requestsViaWebsocket,
-                JsonSerializer = serializer ?? new NewtonsoftJsonSerializer()
-            });
+                UseWebSocketForQueriesAndMutations = requestsViaWebsocket
+            },
+                serializer ?? new NewtonsoftJsonSerializer());
```
sungam3r: Do you have tests for STJ serializer?

rose-a: yes, there is always an abstract base class containing the tests and a specific testclass for each serializer (currently Newtonsoft and `System.Text.Json`)

rose-a:
I think this requires a bump of the major version of this library (as it's a *breaking change*)...

@sungam3r Whats your opinion on this?

sungam3r:
Yes. Personally I am OK with breaking changes.

PR #199, opened 2020-03-22
Add the ability to include Headers for Request Message

nothingalike:
This is how i was able to resolve my issues. I needed to add auth tokens into the headers of the request.

### DIFF
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -145,6 +145,11 @@ namespace GraphQL.Client.Http
                 Content = new StringContent(JsonSerializer.SerializeToString(request), Encoding.UTF8, Options.MediaType)
             };
 
+            foreach(var header in Options.Headers)
+            {
+                message.Headers.Add(header.Key, header.Value);
+            }
+
             if (request is GraphQLHttpRequest httpRequest)
                 httpRequest.PreprocessHttpRequestMessage(message);
 
--- a/src/GraphQL.Client/GraphQLHttpClientOptions.cs
+++ b/src/GraphQL.Client/GraphQLHttpClientOptions.cs
@@ -1,4 +1,5 @@
 using System;
+using System.Collections.Generic;
 using System.Net.Http;
 using System.Net.Http.Headers;
 using System.Threading.Tasks;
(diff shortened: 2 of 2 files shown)


sungam3r:
So it is global (shared) headers for all requests and obviously there may be problems when changing headers in runtime.

rose-a:
Where is the difference to setting the `DefaultRequestHeaders` property of the `HttpClient` used inside `GraphQLHttpClient`?

nothingalike:
From my experience and testing, DefaultRequestHeaders did not add the correct headers to the Http Request.

nothingalike:
@sungam3r  those are probably fair points and open to suggestions, but this update allowed me to execute calls with headers that were needed. More specific: hitting Shopify GraphQL

rose-a:
What exactly did go wrong when you used `DefaultRequestHeaders`? I'd really like to get to the bottom of this, because it's quite hard to believe that this does not work...

nothingalike:
@rose-a so not sure what to tell you other than those are on two different objects. DefaultRequestHeaders are on the actual client itself while the headers im populating go on the RequestMessage themselves. I can rename Header to MessageHeaders. When testing with Fiddler and posting GraphQL messages to Shopify, these were the headers i needed to populate.

rose-a:
> so not sure what to tell you

What exactly does not work when you set the headers you need for the Shopify API using the *DefaultRequestHeaders* property on the *HttpClient* used by *GraphQLHttpClient*?

Maybe show us some code in which the problems manifest themselves?

Citing Microsofts documentation of the *HttpClient.DefaultRequestHeaders Property*:

> The headers which should be sent with each request.

Thats what actually happens and it works for me without any problems (i.e. for sending authentication tokens)!

sungam3r:
My 50 cents - if you look at the source code of HttpClient internals then you will see that `HttpClient.DefaultRequestHeaders` are copied into `HttpRequestMessage.Headers` on each request.

nothingalike:
Interesting ill retry.

nothingalike:
Not sure what i did differently the first time. defaultheaders look good. thank you

arisliang:
Is this added in `master` and published on nuget (latest 4.0.2)?
I couldn't seem to get Option.Headers from nuget 4.0.2. How do we utilise this ability to add headers to request?

Issue #209, opened 2020-04-03
Question about GraphQLHttpRequest.PreprocessHttpRequestMessage


sungam3r:
Why is it needed in `GraphQLHttpRequest` if there is already `GraphQLHttpClientOptions.PreprocessRequest`. Why not have both delegates in `GraphQLHttpClientOptions`? From my point of view it would be a more understandable design rather than mixing delegate inside DTO - GraphQLHttpRequest.

rose-a:
This seems to be the "predecessor" of `GraphQLHttpClientOptions.PreprocessRequest`. 

The only difference makes is that you could do requests with different preprocessing methods, although I can't think of any use case where this would be needed....

So IMO `GraphQLHttpRequest.PreprocessHttpRequestMessage` can be completely eliminated.

sungam3r:
`PreprocessRequest` works with `GraphQLRequest`, `PreprocessHttpRequestMessage` works with `HttpRequestMessage` - different abstraction levels.

> The only difference makes is that you could do requests with different preprocessing methods

True. Well, now I think this is actually useful, as it sets a convenient way to pass parameters without changing the method signature. For example, in this way you can set unique request headers that cannot be set in `DefaultRequestHeaders`. In general, I see asymmetries in the way headers work for request and response. Compare with `GraphQLHttpResponse<T>` : `GraphQLResponse<T>` where we have explicit `ResponseHeaders` property. Perhaps the initial thought was that the `GraphQLHttpRequest` has more initiative and ability to modify the request, so it can obtain the whole `HttpRequestMessage` object to do something with it. 

Need to think more.

rose-a:
`GraphQLHttpRequest.PreprocessHttpRequestMessage` does not work when using a Websocket to send the requests, `GraphQLHttpClientOptions.PreprocessRequest` does...

sungam3r:
Obviously since `GraphQLHttpClientOptions.PreprocessRequest` works with `GraphQLRequest` - DTO, not with `HttpRequestMessage`. I think if in general there is no explicit need for `GraphQLHttpRequest.PreprocessHttpRequestMessage`, then it is better to remove it to simplify design.

rose-a:
#216 is a usecase for `GraphQLHttpRequest.PreprocessHttpRequestMessage`...

sungam3r:
It makes sense.

rose-a:
I assume #218 made this discussion obsolete?

sungam3r:
[...]

PR #218, opened 2020-04-18
Return content on failed http requests

### LINKED ISSUE #87: Not possible to read HttpResponseMessage in case of AggregateException
I'm having a hard time to read the Content from the HttpResponseMessage inside the GraphQLHttpException. The code is throwing AggregateException, which holds the InnerExceptions property of type GraphQLHttpException. When I try to read the Content to figure out why my request is not working .net says the object has been disposed respectively, which is also ok as the client is implemented as a singleton. Why singleton? As far as I understood tthe graphql-client is a wrapper around HttpClient, which should be instantiated as singletons due to some internal bug(s). 

This is what I tried, but no results:

            try
            {

                var request = new GraphQLRequest()
                {
                    Query = CreateQuery(query)
                };

                client.DefaultRequestHeaders.Authorization = new AuthenticationHeaderValue("Bearer","token_goes_here");
                client.Options = new GraphQLClientOptions()
                {
                    MediaType = new MediaTypeHeaderValue("application/json"),
                    EndPoint = new Uri(listingImagesGraphqlUrl, UriKind.Absolute)
                };
                var result = client.PostAsync(request).Result;

                if (result.Errors.Any())
                {
                    var unprocessedIds = result.Errors.Select(x => x.AdditonalEntries)
                        .Where(y => y.ContainsKey("extensions")).Where(z => z.ContainsKey("unprocessedIds"));
                }

            }
            catch (AggregateException exception)
            {
                Logging.WriteError(exception);
                var ex = (GraphQLHttpException) exception.InnerExceptions[0];
                Console.WriteLine(ex.HttpResponseMessage.Content.ReadAsStringAsync());
                throw;
            }

Is it possible to read the response body somehow?

rose-a:
Fixes #87.

Moves the code to cerate the `HttpRequestMessage` from a `GraphQLRequest` info `GraphQLHttpRequest` to allow better customization of the HTTP requests as discussed in #209.


rose-a:
@sungam3r care to review this?

src/GraphQL.Client/GraphQLHttpRequestException.cs
```diff
@@ -0,0 +1,42 @@
...
+        /// <param name="responseHeaders"></param>
+        /// <param name="content"></param>
+        public GraphQLHttpRequestException(HttpStatusCode statusCode, HttpResponseHeaders responseHeaders, string? content) : base($"The HTTP request failed with status code {statusCode}")
+        {
+            StatusCode = statusCode;
+            ResponseHeaders = responseHeaders;
+            Content = content;
+        }
+    }
+
```
sungam3r, line 41:
```suggestion
```

src/GraphQL.Client/GraphQLHttpClient.cs
```diff
@@ -127,32 +127,30 @@ public IObservable<GraphQLResponse<TResponse>> CreateSubscriptionStream<TRespons
         private async Task<GraphQLHttpResponse<TResponse>> SendHttpPostRequestAsync<TResponse>(GraphQLRequest request, CancellationToken cancellationToken = default)
```
sungam3r, line 127: Once you delegated the creation of the `httpRequestMessage` to the caller, this method is no longer `SendHttpPOSTRequestAsync`. It is just `SendHttpRequestAsync` because I can send a GET request in the format defined in [http spec](https://github.com/APIs-guru/graphql-over-http#get).

PR #220, opened 2020-04-19
Fix Map array deserialisation in Newtonsoft serialiser

### LINKED ISSUE #215: GraphQLResponse.Extensions not deserialised properly
I'm trying to include Apollo Tracing data in my responses, which gets included in the Extensions field of the GraphQLResponse object.

For example my query result returns data like in the apollo example at https://github.com/apollographql/apollo-tracing#example

When I look at the values in the extensions map, it doesn't seem to have deserialised the trace extension object correctly.

`extensions.tracing.parsing` and similar fields are fine, but `extensions.tracing.execution.resolvers` is always an empty array.

This seems to be a problem in `GraphQL.Client.Serializer.Newtonsoft.MapConverter.ReadArray`, because the items in the resolvers array are of type `JTokenType.Property` and are therefore skipped / ignored.

Is it possible to extend the deserialiser to support this structure of extension JSON?

PaulKlein:
Resolves #215

### DIFF
--- a/src/GraphQL.Client.Serializer.Newtonsoft/MapConverter.cs
+++ b/src/GraphQL.Client.Serializer.Newtonsoft/MapConverter.cs
@@ -51,7 +51,7 @@ namespace GraphQL.Client.Serializer.Newtonsoft
             return result;
         }
 
-        private IEnumerable<object> ReadArray(JToken element)
+        private IEnumerable<object> ReadArray(JArray element)
         {
             foreach (var item in (JArray)element)
             {
--- a/src/GraphQL.Client.Serializer.Newtonsoft/MapConverter.cs
+++ b/src/GraphQL.Client.Serializer.Newtonsoft/MapConverter.cs
@@ -53,7 +53,7 @@ namespace GraphQL.Client.Serializer.Newtonsoft
 
         private IEnumerable<object> ReadArray(JToken element)
         {
-            foreach (var item in element.Values())
+            foreach (var item in (JArray)element)
             {
                 if (IsUnsupportedJTokenType(item.Type))
                     continue;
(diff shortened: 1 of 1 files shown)


PaulKlein:
Sorry, for some reason I got the issue number way wrong when committing this but I'm not sure if I can change it.

I am also not sure what's causing the failing test - any ideas?

rose-a:
Did the tests pass locally? There is a persistent issue with the integration tests where some tests randomly fail on GitHub Actions...

src/GraphQL.Client.Serializer.Newtonsoft/MapConverter.cs
```diff
@@ -53,7 +53,7 @@ token switch
 
         private IEnumerable<object> ReadArray(JToken element)
```
rose-a, line 54:
```suggestion
        private IEnumerable<object> ReadArray(JArray element)
```

src/GraphQL.Client.Serializer.Newtonsoft/MapConverter.cs
```diff
@@ -53,7 +53,7 @@ token switch
...
         {
-            foreach (var item in element.Values())
+            foreach (var item in (JArray)element)
```
rose-a, line 56:
```suggestion
            foreach (var item in element)
```

rose-a:
Tests pass on my development machine, so nothing in the way of merging your modifications... 

Thanks for digging into this!

PaulKlein:
Thanks for taking a look at this so quickly.

Yes, your changes tidy this up even more, thanks for the feedback there.
The tests do run fine on my machine as well

Thanks again

rose-a:
Included in v3.0.3 😉

PR #225, opened 2020-04-21
Fix inconsistency in mapconverters

### LINKED ISSUE #223: Inconsistent MapConverter array return values between NewtonSoft and SystemText Json Serializers
If a GraphQL errors collection contains extensions where a map value is an array, the NewtonSoft MapConverter returns an `IEnumerable<object>`, where the SystemTextJson MapConverter returns a `List<object>`.  These should both work the same way.

While both MapConverter implementations of ReadArray() return an `IEnumerable<object>`, the calling methods differ in that the NewtonSoft implementation of ReadToken simply passes the IEnumerable along, while the SystemTextJson implementation of ReadValue calls ToList() on the result of ReadArray().

I suggest these behave consistently (I would vote for adding ToList() to the NewtonSoft MapConverter implementation), although either way can work.

An example that can be used to reproduce would be as simple as
```json
"extensions": {
        "codes": [
          "some code"
        ]
}
```
and deserializing using both the NewtonSoft and SystemTextJson MapConverters.

rose-a:
Resolves #223

### DIFF
--- a/src/GraphQL.Client.Serializer.Newtonsoft/MapConverter.cs
+++ b/src/GraphQL.Client.Serializer.Newtonsoft/MapConverter.cs
@@ -1,6 +1,6 @@
 using System;
 using System.Collections.Generic;
-using System.Runtime.InteropServices;
+using System.Linq;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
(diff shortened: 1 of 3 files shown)


BillBaird:
Thanks for the quick turnaround!

PR #226, opened 2020-04-21
Non generic interface for GraphQLResponse

### LINKED ISSUE #224: Can GraphQLResponse<T> implement an IGraphQLResponse non-generic interface?
With changes from 2.0 `GraphQLResponse` to 3.0 `GraphQLResponse<T>`, common error handling logic must use generics even though the Errors and Extensions have no need for the generic type T.  In my case, I have various common client error handling logic that had to be unnecessarily extended to work with generics.

I propose something like:
```csharp
public interface IGraphQLReadOnlyResponse
{
    public object Data { get; }
    public GraphQLError[]? Errors { get; }
    public Map? Extensions { get; }
}
```

I'm less interested in the `Data` property as an object (explicit interface implementation) than I am for `Errors` and `Extensions`.  Providing a interface which allows these to be used by utility code that didn't know or care about, or even have visibility to the type T would allow the utility code to provide a consistent way of handling the errors simply by passing the `GraphQLResponse` to the error handler without regard to the generic type T.

For instance:
```csharp
public void HandleErrors(IGraphQLReadOnlyResponse response)
{
    if (response.Errors ...)
}
...
var response = await client.SendQueryAsync<MyType>( ... );
HandleErrors(response);
```

Can this simple change be included?  It would allow me to remove unnecessary complexity from client logic.

rose-a:
Resolves #224

### DIFF
--- a/GraphQL.Client.sln
+++ b/GraphQL.Client.sln
@@ -33,9 +33,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "GraphQL.Server.Test", "test
 EndProject
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "workflows", "workflows", "{05CAF9B2-981E-40C0-AE31-5FA56E351F12}"
 	ProjectSection(SolutionItems) = preProject
-		.github\workflows\branches-ubuntu.yml = .github\workflows\branches-ubuntu.yml
 		.github\workflows\branches.yml = .github\workflows\branches.yml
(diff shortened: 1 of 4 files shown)


BillBaird:
Thanks for the quick turnaround!

yaynick:
Oh man had already finished refactoring half the codebase for cleaner error handling due to the  generics, only to desperately give up and check GH issues in search of a better way.

Thank you so much @rose-a for the amazing turnaround and thank you @BillBaird for having suggested it!

PR #252, opened 2020-06-17
Create new GraphQLHttpWebSocket instance if needed

Maydayof:
Create new GraphQLHttpWebSocket instance if the UseWebSocketForQueriesAndMutations is true

### DIFF
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -188,8 +188,8 @@ namespace GraphQL.Client.Http
                 Debug.WriteLine($"Disposing GraphQLHttpClient on endpoint {Options.EndPoint}");
                 _cancellationTokenSource.Cancel();
                 HttpClient.Dispose();
-                if ( (bool)(_lazyHttpWebSocket?.IsValueCreated) )
-                    _lazyHttpWebSocket?.Value.Dispose();
+                if ( _lazyHttpWebSocket.IsValueCreated )
+                    _lazyHttpWebSocket.Value.Dispose();
                 _cancellationTokenSource.Dispose();
             }
         }
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -188,7 +188,8 @@ namespace GraphQL.Client.Http
                 Debug.WriteLine($"Disposing GraphQLHttpClient on endpoint {Options.EndPoint}");
                 _cancellationTokenSource.Cancel();
                 HttpClient.Dispose();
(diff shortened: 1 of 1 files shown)


src/GraphQL.Client/GraphQLHttpClient.cs
```diff
@@ -63,7 +63,8 @@ public GraphQLHttpClient(GraphQLHttpClientOptions options, IGraphQLWebsocketJson
             if (!HttpClient.DefaultRequestHeaders.UserAgent.Any())
                 HttpClient.DefaultRequestHeaders.UserAgent.Add(new ProductInfoHeaderValue(GetType().Assembly.GetName().Name, GetType().Assembly.GetName().Version.ToString()));
 
-            _graphQlHttpWebSocket = new GraphQLHttpWebSocket(GetWebSocketUri(), this);
+            if (options.UseWebSocketForQueriesAndMutations)
```
rose-a, line 66: Subscriptions must always use the websocket connection (obviously they don't work with regular HTTP requests). 

You need to make sure that an instance of `GraphQLWebSocket` is created when a subscription is created, otherwise you will receive a null reference exception on creating a subscription.

Maydayof:
To instantiate a GraphQLHttpWebSocket requires a call into an instance method and pass an instance reference (/* public GraphQLHttpWebSocket(Uri webSocketUri, GraphQLHttpClient client) */ )
But, a field initializer cannot reference non-static field, method, property. At the moment I just can't figure out, how should I use the Lazy<T> variant? Any idea?

rose-a:
I think its well described in the [API Reference](https://docs.microsoft.com/en-us/dotnet/api/system.lazy-1?view=netcore-3.1).

Create a private field for a `Lazy<GraphQLHttpWebSocket>` and assign it inside the constructor using the factory method `() => new GraphQLHttpWebSocket(GetWebSocketUri(), this)`.

Then you access the actual instance of `GraphQLHttpWebSocket` using `lazyWebsocket.Value`.
The instance will then be created on fist access to `lazyWebsocket.Value`.

Maydayof:
Thx, I have moved the initialization into the ctor, that was the problem.

src/GraphQL.Client/GraphQLHttpClient.cs
```diff
@@ -196,7 +188,7 @@ protected virtual void Dispose(bool disposing)
                 Debug.WriteLine($"Disposing GraphQLHttpClient on endpoint {Options.EndPoint}");
                 _cancellationTokenSource.Cancel();
                 HttpClient.Dispose();
-                _graphQlHttpWebSocket?.Dispose();
+                _lazyHttpWebSocket?.Value.Dispose();
```
rose-a, line 191:
```suggestion
                if(_lazyHttpWebSocket?.HasValue) _lazyHttpWebSocket?.Value.Dispose();
```

Otherwise we'll create an instance on calling `Dispose()` 😉

Issue #262, opened 2020-08-16
GraphQL.Client: add Websocket support for Blazor WebAssembly host


bjorg:
I would like to help address this issue as I'm a big fan of _Blazor WebAssembly_ and a strong believer in GraphQL as the best API tech. So, supporting `GraphQL.Client` is of paramount interest. If there is already an effort underway to achieve this, please close this issue as a duplicate and point me in the right direction.

# Problem
I've started looking at the code to better understand where the problem is originating from. At the core of the problem are the limited capabilities of the Blazor WASM runtime. For instance, it's not possible to create threads or even block on a thread with `Task.Wait()` or `Task.Result`. All operations must be asynchronous as there is only one main thread in the browser execution environment.

The implementation of the `GraphQL.Client.Http.Websocket.GraphQLHttpWebSocket` class uses [System.Reactive.Concurency.EventLoopScheduler](https://github.com/dotnet/reactive/blob/master/Rx.NET/Source/src/System.Reactive/Concurrency/EventLoopScheduler.cs), which creates a thread internally. This causes the `System.NotSupportedException: Cannot start threads on this runtime.` error when the websocket connection is opened.

# Solution
There are 2 approaches to potentially address this problem:

## Approach 1 - Blazor WASM-friendly `System.Reactive.Concurrency.IScheduler` implementation
Implement a Blazor WASM-friendly version of `System.Reactive.Concurrency.IScheduler` that does not require background threads. It would only work reliably when scheduled operations are asynchronous. However, it would be the least invasive change.

**Benefits**
* Least invasive
* Least amount of new code
* Requires small initialization change in `GraphQLHttpWebSocket` that could be controlled via a `GraphQLHttpClientOptions` property to allow current and new implementations to exist side-by-side

**Drawback**
* No compile-time detection if the asynchronous requirement is not honored

## Approach 2 - Replace `System.Reactive` in favor of a pure asynchronous implementation

**Benefits**
* Removes dependency on `System.Reactive` (less is better)
* Clean asynchronous interface enables compiler to assist with `async/await` code pattern enforcement.
* Doesn't require to understand how `System.Reactive` works by deferring continuations to `Task`-related methods

**Drawbacks**
* Breaking change if public interface of `GraphQL.Client` exposes `IObservable`
* Lots of code changes
* Even more complexity if old and new implementations need to exists side-by-side

# Additional

As follow up to this issue, there should also be a sample Blazor WASM application to showcase and test the proper behavior.

# Related
[...]

PR #274, opened 2020-09-06
Blazor WebAssembly support: added try-catch for unsupported properties

bjorg:
Addresses remaining compatibility issues for Blazor WebAssembly support (see #262)

### DIFF
--- a/src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
+++ b/src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
@@ -619,7 +619,7 @@ namespace GraphQL.Client.Http.Websocket
 
         /// <summary>
         /// Task to await the completion (a.k.a. disposal) of this websocket.
-        /// </summary>
+        /// </summary> 
         /// Async disposal as recommended by Stephen Cleary (https://blog.stephencleary.com/2013/03/async-oop-6-disposal.html)
         public Task? Completion { get; private set; }
 
@@ -643,7 +643,7 @@ namespace GraphQL.Client.Http.Websocket
             _exceptionSubject?.OnCompleted();
             _exceptionSubject?.Dispose();
             _internalCancellationTokenSource.Dispose();
-
+            
             Debug.WriteLine("GraphQLHttpWebSocket disposed");
         }
 
--- a/src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
+++ b/src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
@@ -407,8 +407,22 @@ namespace GraphQL.Client.Http.Websocket
 #else
                 _clientWebSocket = new ClientWebSocket();
                 _clientWebSocket.Options.AddSubProtocol("graphql-ws");
-                _clientWebSocket.Options.ClientCertificates = ((HttpClientHandler)Options.HttpMessageHandler).ClientCertificates;
-                _clientWebSocket.Options.UseDefaultCredentials = ((HttpClientHandler)Options.HttpMessageHandler).UseDefaultCredentials;
+                try
+                {
+                    _clientWebSocket.Options.ClientCertificates = ((HttpClientHandler)Options.HttpMessageHandler).ClientCertificates;
+                }
+                catch(PlatformNotSupportedException)
+                {
+                    Debug.WriteLine("unable to set Options.ClientCertificates property; platform does not support it");
+                }
+                try
+                {
+                    _clientWebSocket.Options.UseDefaultCredentials = ((HttpClientHandler)Options.HttpMessageHandler).UseDefaultCredentials;
+                }
+                catch(PlatformNotSupportedException)
+                {
+                    Debug.WriteLine("unable to set Options.UseDefaultCredentials property; platform does not support it");
+                }
                 Options.ConfigureWebsocketOptions(_clientWebSocket.Options);
 #endif
                 return _initializeWebSocketTask = ConnectAsync(_internalCancellationToken);
@@ -605,7 +619,7 @@ namespace GraphQL.Client.Http.Websocket
 
         /// <summary>
         /// Task to await the completion (a.k.a. disposal) of this websocket.
-        /// </summary> 
+        /// </summary>
         /// Async disposal as recommended by Stephen Cleary (https://blog.stephencleary.com/2013/03/async-oop-6-disposal.html)
         public Task? Completion { get; private set; }
 
@@ -629,7 +643,7 @@ namespace GraphQL.Client.Http.Websocket
             _exceptionSubject?.OnCompleted();
             _exceptionSubject?.Dispose();
             _internalCancellationTokenSource.Dispose();
-            
+
             Debug.WriteLine("GraphQLHttpWebSocket disposed");
         }
 
--- a/src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
+++ b/src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
@@ -411,7 +411,7 @@ namespace GraphQL.Client.Http.Websocket
                 {
                     _clientWebSocket.Options.ClientCertificates = ((HttpClientHandler)Options.HttpMessageHandler).ClientCertificates;
                 }
-                catch(PlatformNotSupportedException)
+                catch (PlatformNotSupportedException)
                 {
                     Debug.WriteLine("unable to set Options.ClientCertificates property; platform does not support it");
                 }
@@ -419,7 +419,7 @@ namespace GraphQL.Client.Http.Websocket
                 {
                     _clientWebSocket.Options.UseDefaultCredentials = ((HttpClientHandler)Options.HttpMessageHandler).UseDefaultCredentials;
                 }
-                catch(PlatformNotSupportedException)
+                catch (PlatformNotSupportedException)
                 {
                     Debug.WriteLine("unable to set Options.UseDefaultCredentials property; platform does not support it");
                 }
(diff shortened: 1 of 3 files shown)


src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
```diff
@@ -407,8 +407,22 @@ public Task InitializeWebSocket()
 #else
                 _clientWebSocket = new ClientWebSocket();
                 _clientWebSocket.Options.AddSubProtocol("graphql-ws");
-                _clientWebSocket.Options.ClientCertificates = ((HttpClientHandler)Options.HttpMessageHandler).ClientCertificates;
-                _clientWebSocket.Options.UseDefaultCredentials = ((HttpClientHandler)Options.HttpMessageHandler).UseDefaultCredentials;
+                try
+                {
+                    _clientWebSocket.Options.ClientCertificates = ((HttpClientHandler)Options.HttpMessageHandler).ClientCertificates;
+                }
+                catch(PlatformNotSupportedException)
```
sungam3r, line 414:
```suggestion
                catch (PlatformNotSupportedException)
```

src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
```diff
@@ -407,8 +407,22 @@ public Task InitializeWebSocket()
...
+                {
+                    Debug.WriteLine("unable to set Options.ClientCertificates property; platform does not support it");
+                }
+                try
+                {
+                    _clientWebSocket.Options.UseDefaultCredentials = ((HttpClientHandler)Options.HttpMessageHandler).UseDefaultCredentials;
+                }
+                catch(PlatformNotSupportedException)
```
sungam3r, line 422:
```suggestion
                catch (PlatformNotSupportedException)
```

sungam3r:
Just a thought. Could/should we exclude these properties from compilation for different platforms (multitargeting). Silently catching exceptions may be not optimal. Call to `Debug.WriteLine` will be excluded from release build.

Shane32:
> Just a thought. Could/should we exclude these properties from compilation for different platforms (multitargeting). Silently catching exceptions may be not optimal. Call to `Debug.WriteLine` will be excluded from release build.

I might not exclude them, as they are not excluded from the targeted platform, and it complicates the build process.  Perhaps future versions of the platform will support it.  But I agree that silently catching exceptions isn't optimal.

Here's another idea: perhaps only set these properties if they were set in the options class.  Then, if a person was to trying to set one of these properties, they would property receive a `PlatformNotSupported` exception when attempting to execute the query.  I've not looked at the code in-depth so maybe that doesn't work or isn't applicable.

sungam3r:
This will work if the getters for these properties don't throw PNSE. Then you can compare the values and call setters only if values differ.

rose-a:
> Here's another idea: perhaps only set these properties if they were set in the options class.

I'd prefer this approach, too.

bjorg:
I'll investigate what can be done today. First, I'll check if the getter can be used. If so, I'll use a comparison to set it. Otherwise, I'll add a field to the options class to track when a property is set and key off that.

bjorg:
The getter fails on `((HttpClientHandler)Options.HttpMessageHandler).ClientCertificates` (ditto for `UseDefaultCredentials`). So it's not even possible to read an option that was set by the user. That said, how would user ever set it in the first place? This makes me think the first approach of swallowing the exception may be the way to go. Thoguhts?

src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
bjorg: done

src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
bjorg: done

Shane32:
What if UseDefaultCredentials was a nullable boolean and only sets the property if it HasValue? And similarly only set the other property if not null?

bjorg:
@Shane32 Unfortunately, `ClientCertificates` and `UseDefaultCredentials` are properties on the built-in `System.Net.Http.HttpClientHandler` class. Otherwise, I would have added a boolean field to track if they were modified.

bjorg:
Pushed a few more changes:
1) Preserve query parameters in endpoint URI (this is required for authenticating with [AppSync](https://docs.aws.amazon.com/appsync/latest/devguide/real-time-websocket-client.html#header-parameter-format-based-on-appsync-api-authorization-mode))
2) Allow end-point URI to begin with `wss://` or `ws://`. Unfortunately, AppSync uses a different hostname for their REST vs. WebSocket API endpoints. Being able to specify the `wss://` endpoint directly will make life a bit easier.
3) Automatically set `Options.UseWebSocketForQueriesAndMutations` to `true` when the endpoint scheme is either `wss://` or `ws://`.

src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
```diff
@@ -407,8 +407,22 @@ public Task InitializeWebSocket()
...
                 _clientWebSocket.Options.AddSubProtocol("graphql-ws");
-                _clientWebSocket.Options.ClientCertificates = ((HttpClientHandler)Options.HttpMessageHandler).ClientCertificates;
-                _clientWebSocket.Options.UseDefaultCredentials = ((HttpClientHandler)Options.HttpMessageHandler).UseDefaultCredentials;
+                try
+                {
+                    _clientWebSocket.Options.ClientCertificates = ((HttpClientHandler)Options.HttpMessageHandler).ClientCertificates;
+                }
+                catch (PlatformNotSupportedException)
+                {
+                    Debug.WriteLine("unable to set Options.ClientCertificates property; platform does not support it");
```
sungam3r, line 416: It is noop for release build.

sungam3r:
Let's not mix different fixes in one PR.

src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
bjorg: I know. I figured it was better than just a comment.

sungam3r:
> This makes me think the first approach of swallowing the exception may be the way to go. 

Yes, possible. You just need to clearly reflect this issue in the documentation/faq.

bjorg:
> Yes, possible. You just need to clearly reflect this issue in the documentation/faq.

Ok. Which doc do I need to amend?

bjorg:
> Let's not mix different fixes in one PR.

Reverted. Will file new issue and submit new PR.

Shane32:
> @Shane32 Unfortunately, `ClientCertificates` and `UseDefaultCredentials` are properties on the built-in `System.Net.Http.HttpClientHandler` class. Otherwise, I would have added a boolean field to track if they were modified.

I see.

> > This makes me think the first approach of swallowing the exception may be the way to go.
> 
> Yes, possible. You just need to clearly reflect this issue in the documentation/faq.

Agreed

sungam3r:
> Ok. Which doc do I need to amend?

Readme.md

src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
sungam3r: Probably `Console.WriteLine` (stdout) will be even better.

bjorg: That seems worse as it would interfere with console apps. No production code should ever have any `Console.WriteLine()` statements in it.

bjorg:
Added a note on Blazor WebAssembly limitations to _ReadMe.md_.

src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
Shane32: Well that’s absolutely true for desktop apps. But we should only get the PlatformNotSupported exception in the web where it’s appropriate to display informational logs via the “console“...

bjorg: So let me see if I understand the desired behavior.

**Scenario 1:** The developer does not set these properties, but the code always logs anyway to the web console that it was unable to copy them to `ClientWebSocket` instance.

**Scenario 2:** The developer does attempt to set these properties on the [System.Net.Http.HttpClientHandler](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclienthandler?view=netcore-3.1) instance via `GraphQLHttpClient.Options.HttpMessageHandler` and immediately gets a `PlatformNotSupportedException` error, thus never reaching the code in the try-catch statement and therefore never seeing the web console log message in the first place.

Did I miss something? Because I fail to see when it is useful to log to the web console.

Shane32: I see. Agreed there’s no need for console logs.

Shane32:
> The getter fails on `((HttpClientHandler)Options.HttpMessageHandler).ClientCertificates` (ditto for `UseDefaultCredentials`). So it's not even possible to read an option that was set by the user. That said, how would user ever set it in the first place? This makes me think the first approach of swallowing the exception may be the way to go. Thoguhts?

I missed this earlier. Looks like you’ve got it all figured out! 👍

src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
sungam3r: 😄

bjorg:
Let me know if anything else is needed to merge this change.

Shane32 (approved):

sungam3r (approved):

rose-a (approved):

rose-a:
Thanks @bjorg for your contribution! And thanks to @Shane32 and @sungam3r for supporting him ;)

PR #276, opened 2020-09-08
Allow wss:// scheme, preserve uri query parameters

### LINKED ISSUE #275: Endpoint scheme and query parameters are ignored when opening websocket connection
The `GraphQLHttpClient.GetWebSocketUri()` method doesn't apply the query parameters of the endpoint URI. In addition, the endpoint scheme is incorrectly set when it's already `wss://`.

The method should be changed to:
```csharp
private Uri GetWebSocketUri()
{
    string webSocketSchema = Options.EndPoint.Scheme == "https"
        ? "wss"
        : Options.EndPoint.Scheme == "http"
        ? "ws"
        : Options.EndPoint.Scheme;
    return new Uri($"{webSocketSchema}://{Options.EndPoint.Host}:{Options.EndPoint.Port}{Options.EndPoint.AbsolutePath}{Options.EndPoint.Query}");
}
```

bjorg:
Fixes #275 - Endpoint scheme and query parameters are ignored when opening websocket connection

I also added the change to automatically switch `Options.UseWebSocketForQueriesAndMutations` to `true` when the endpoint scheme is a websocket.


bjorg:
Closing this PR for now until I figure out everything that is needed to make AppSync work.

PR #277, opened 2020-09-15
handle WebSocket url scheme

### LINKED ISSUE #275: Endpoint scheme and query parameters are ignored when opening websocket connection

rose-a:
closes #275

### DIFF
--- a/src/GraphQL.Client/UriExtensions.cs
+++ b/src/GraphQL.Client/UriExtensions.cs
@@ -22,7 +22,7 @@ namespace GraphQL.Client.Http
                 return uri;
 
             string webSocketScheme = uri.Scheme == "https" ? "wss" : "ws";
-            return new Uri($"{webSocketScheme}://{uri.Host}:{uri.Port}{uri.AbsolutePath}");
+            return new Uri($"{webSocketScheme}://{uri.Host}:{uri.Port}{uri.PathAndQuery}");
         }
     }
 }
--- a/src/GraphQL.Client/UriExtensions.cs
+++ b/src/GraphQL.Client/UriExtensions.cs
@@ -1,0 +1,28 @@
+﻿using System;
+
+namespace GraphQL.Client.Http
+{
+    public static class UriExtensions
+    {
+        /// <summary>
+        /// Returns true if <see cref="Uri.Scheme"/> equals "wss" or "ws"
+        /// </summary>
+        /// <param name="uri"></param>
+        /// <returns></returns>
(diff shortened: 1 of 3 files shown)


src/GraphQL.Client/UriExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace GraphQL.Client.Http
+{
+    public static class UriExtensions
+    {
+        /// <summary>
+        /// Returns true if <see cref="Uri.Scheme"/> equals "wss" or "ws"
+        /// </summary>
+        /// <param name="uri"></param>
+        /// <returns></returns>
+        public static bool HasWebSocketScheme(this Uri uri) => uri.Scheme.Equals("wss") || uri.Scheme.Equals("ws");
```
bjorg, line 12: This should be a case-insensitive comparison since `"HTTPS"` is just as valid as `"https"` for the scheme.

src/GraphQL.Client/UriExtensions.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// Infers the websocket uri from <paramref name="uri"/>.
+        /// </summary>
+        /// <param name="uri"></param>
+        /// <returns></returns>
+        public static Uri GetWebSocketUri(this Uri uri)
+        {
+            if (uri.HasWebSocketScheme())
+                return uri;
+
+            string webSocketScheme = uri.Scheme == "https" ? "wss" : "ws";
```
bjorg, line 24: Comparison needs to be case-insensitive. Also, why default to `"ws"` when the current scheme is not `"https"`? For example, `"ftp"` is a valid URI scheme, but it should not be converted to `"ws"`. Instead, I would expect this method to throw an exception when it is not a HTTP(S) scheme.

rose-a: done it properly in #279

Issue #280, opened 2020-09-16
Blazor WebAssembly - handle unsupported properties properly


rose-a:
In #274 we've added a workaround for unsupported properties in WASM which swallows the `PlatformNotSupported` exceptions.
We should add proper guards to avoid execution this part of the code completely when running on WebAssembly.

@bjorg, I've found [this answer](https://github.com/dotnet/aspnetcore/issues/18268#issuecomment-573123617) in a GitHub issue on "detecting if the code runs in WASM".

Could you please try if `RuntimeInformation.IsOSPlatform(OSPlatform.Create("WEBASSEMBLY"))` correctly detects your code running in WebAssembly? This way we could purposely avoid accessing these properties without swallowing exceptions...

Seems that from .NET 5 on we'll get a `OperatingSystem.IsBrowser()` method to do just this, according to [this article](https://devblogs.microsoft.com/dotnet/the-future-of-net-standard/#dealing-with-apis-that-are-unsupported-in-blazor-webassembly)...

bjorg:
I saw the announcement for `OperatingSystem.IsBrowser()`, but I was not aware of `RuntimeInformation.IsOSPlatform(OSPlatform.Create("WEBASSEMBLY"))`.

I'm happy to test it and see if it works as expected. The expectation being that unsupported properties are simply ignored and not set, correct?

rose-a:
yeah, I'm thinking something like

```csharp
if(!RuntimeInformation.IsOSPlatform(OSPlatform.Create("WEBASSEMBLY"))){
    _clientWebSocket.Options.ClientCertificates = ((HttpClientHandler)Options.HttpMessageHandler).ClientCertificates;
    _clientWebSocket.Options.UseDefaultCredentials = ((HttpClientHandler)Options.HttpMessageHandler).UseDefaultCredentials;
}
```

bjorg:
Here you go: https://github.com/graphql-dotnet/graphql-client/pull/283

Please verify that I didn't break anything for other runtime targets.

PR #283, opened 2020-09-17
check if running in WebAssembly instead of catching PlatformNotSupportedException

bjorg:
This pull-request fixes https://github.com/graphql-dotnet/graphql-client/issues/280

I used explicit class names for `RuntimeInformation` and `OSPlatform` instead of adding a using statement that would only apply conditionally. Let me know which style you prefer.

There was also a trailing space that was automatically nuked by my editor, which I didn't catch before pushing my changes. Let me know if you'd like it back.

I tested this on a Blazor WebAssembly app built with .NET Core 3.1.402. I did NOT test it on another target.


sungam3r (approved):

rose-a (approved):
I ran the code under netcoreapp2.1, netcoreapp3.1 and net471.

netcoreapp2.1 and netcoreapp3.1 work as expected, while net471 uses the library built for net461, which bypasses this code.

Since the .NET Framework is not relevant for WebAssembly I'd say we're safe.

PR #285, opened 2020-09-28
Adding initialization headers to a WebSocket GraphQL client

codewalkerdan:
Not just authorization, but other headers could be added that may be required when initializing a web socket connection.
It supports adding authorization tokens

### DIFF
--- a/src/GraphQL.Client.Abstractions.Websocket/GraphQLWebSocketRequest.cs
+++ b/src/GraphQL.Client.Abstractions.Websocket/GraphQLWebSocketRequest.cs
@@ -34,9 +34,9 @@ namespace GraphQL.Client.Abstractions.Websocket
         /// <summary>
         /// The payload of the websocket request
         /// </summary>
-        public GraphQLRequest Payload
+        public object Payload
         {
-            get => ContainsKey(PAYLOAD_KEY) ? (GraphQLRequest)this[PAYLOAD_KEY] : null;
+            get => ContainsKey(PAYLOAD_KEY) ? this[PAYLOAD_KEY] : null;
             set => this[PAYLOAD_KEY] = value;
         }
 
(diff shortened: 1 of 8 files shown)


src/GraphQL.Client.Abstractions.Websocket/GraphQLWebSocketRequest.cs
```diff
@@ -34,9 +34,9 @@ public string Type
         /// <summary>
         /// The payload of the websocket request
         /// </summary>
-        public GraphQLRequest Payload
+        public object Payload
```
sungam3r, line 37: breaking change

src/GraphQL.Primitives/GraphQL.Primitives.csproj
```diff
@@ -6,6 +6,7 @@
     <Description>GraphQL basic types</Description>
     <RootNamespace>GraphQL</RootNamespace>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <Version>1.0.0</Version>
```
sungam3r, line 9: ?

src/GraphQL.Client/GraphQL.Client.csproj
```diff
@@ -5,6 +5,14 @@
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
     <RootNamespace>GraphQL.Client.Http</RootNamespace>
+    <Authors>Deinok,Alexander Rose,graphql-dotnet</Authors>
```
sungam3r, line 8: https://github.com/graphql-dotnet/graphql-client/blob/master/root.props

sungam3r (changes requested):
These edits don't seem right to me.

rose-a:
@codewalkerdan Thanks for your contribution. I'm closing this in favor of #294, which basically does the same thing albeit keeping types and names close to the [Apollo GraphQL over WebSocket Protocol Spec](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_connection_init), which this library aims to fully implement.

codewalkerdan:
Sorry for not replying. I have little time, but thank you for the feedback.

PR #291, opened 2020-10-06
GraphQLRequest fixes

bjorg:
1. Preserve all entries in `GraphQLRequest` when using the copy constructor.
2. Avoid `NullReferenceException` on `GetHashCode()` when `Query` is not set.

### DIFF
--- a/src/GraphQL.Primitives/GraphQLRequest.cs
+++ b/src/GraphQL.Primitives/GraphQLRequest.cs
@@ -49,6 +49,14 @@ namespace GraphQL
             OperationName = operationName;
         }
 
+        public GraphQLRequest(IEnumerable<KeyValuePair<string, object>> values)
+        {
+            foreach(var kv in values)
+            {
+                Add(kv.Key, kv.Value);
+            }
+        }
+
         /// <summary>
         /// Returns a value that indicates whether this instance is equal to a specified object
         /// </summary>
@@ -86,7 +94,7 @@ namespace GraphQL
         {
             unchecked
             {
-                var hashCode = Query.GetHashCode();
+                var hashCode = Query?.GetHashCode() ?? 0;
                 hashCode = (hashCode * 397) ^ OperationName?.GetHashCode() ?? 0;
                 hashCode = (hashCode * 397) ^ Variables?.GetHashCode() ?? 0;
                 return hashCode;
--- a/src/GraphQL.Primitives/GraphQLRequest.cs
+++ b/src/GraphQL.Primitives/GraphQLRequest.cs
@@ -84,16 +84,7 @@ namespace GraphQL
         /// <summary>
         /// <inheritdoc cref="object.GetHashCode"/>
         /// </summary>
-        public override int GetHashCode()
-        {
-            unchecked
-            {
-                var hashCode = Query?.GetHashCode() ?? 0;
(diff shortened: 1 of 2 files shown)


src/GraphQL.Primitives/GraphQLRequest.cs
```diff
@@ -49,6 +49,14 @@ public GraphQLRequest(string query, object? variables = null, string? operationN
             OperationName = operationName;
         }
 
+        public GraphQLRequest(IEnumerable<KeyValuePair<string, object>> values)
```
Shane32, line 52: This constructor seems a bit odd, but so does the fact that `GraphQLRequest` inherits from a `Dictionary<>`...

I would suggest this change:
```suggestion
        public GraphQLRequest(GraphQLRequest other)
```
The `foreach` code below should still be valid.

Shane32: If someone actually needed to clone a generic dictionary into a `GraphQLRequest` class, I'd say they write a `foreach` snippet themselves, but this would not be typical.  Correct???  I've only used this library a few times, so lmk if not.

bjorg: Sounds reasonable. You're right that `IEnumerable<KeyValuePair<string, object>>` is more generic than it needs to be.

bjorg: Fixed copy constructor.

Shane32 (approved):

src/GraphQL.Primitives/GraphQLRequest.cs
```diff
@@ -49,6 +49,14 @@ public GraphQLRequest(string query, object? variables = null, string? operationN
             OperationName = operationName;
         }
 
+        public GraphQLRequest(GraphQLRequest other)
```
rose-a, line 52: I'd prefer `public GraphQLRequest(GraphQLRequest other) : base(other)`, since `Dictionary<string,object>` already has a "copy-constructor"

rose-a (changes requested):

src/GraphQL.Primitives/GraphQLRequest.cs
```diff
@@ -86,7 +94,7 @@ public override int GetHashCode()
         {
             unchecked
             {
-                var hashCode = Query.GetHashCode();
+                var hashCode = Query?.GetHashCode() ?? 0;
```
sungam3r, line 97: public override int GetHashCode() => (Query, OperationName, Variables).GetHashCode();

src/GraphQL.Primitives/GraphQLRequest.cs
bjorg: Much better indeed!

src/GraphQL.Primitives/GraphQLRequest.cs
bjorg: That's a neat trick! Are you okay with ignoring all the other keys in the dictionary though? Obviously the original implementation did.

sungam3r: I am OK with any options. I just don't like these hand-written methods for a long time.

rose-a (approved):

Issue #293, opened 2020-10-07
Fix implementation details of the GraphQL over WebSocket Protocol


rose-a:
https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md

Particular Issues:
- *Connected Phase* not honored, connection init is sent on each subscription and never on queries or mutations via websocket
- `GQL_CONNECTION_ACK` is ignored
- `GQL_CONNECTION_TERMINATE` is not sent before closing the websocket connection

PR #294, opened 2020-10-07
Add websocket ConnectionParams object 

### LINKED ISSUE #250: Connecting websocket to apollo server
I am trying to create a subscription to my apollo server running on node.js.

I am getting the error: "binary websocket messages are not supported" which I see is in the GraphQLHttpWebSocket class.

Does anyone have any experience using graphql-client to connect a websocket to an apollo server, or what configuration would be needed to make it connect correctly?

### LINKED ISSUE #269: Set header on web socket requests
Hello, 
I'm trying to set a header on the ws requests in order to authenticate the subscriptions, but I cant seem to find a way to do so.
My current code:

```
HttpClient = new HttpClient();
HttpClient.DefaultRequestHeaders.Add("Authorization", $"Bearer MY TOKEN");

var GQLClient = new GraphQLHttpClient(
  new GraphQLHttpClientOptions
  {
    EndPoint = new Uri("MY URI"),
    UseWebSocketForQueriesAndMutations = false,
    OnWebsocketConnected = OnWebSocketConnect,
  },
  new NewtonsoftJsonSerializer(),
  HttpClient);

var ws = new ClientWebSocket();;
ws.Options.SetRequestHeader("Authorization", $"Bearer MY TOKEN");

GQLClient.Options.ConfigureWebsocketOptions(ws.Options);
```

On the apollo server no headers are found, can anyone help?

rose-a:
This is specified in the [Apollo GraphQL over WebSocket Protocol specification](https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_connection_init) and there is currently no way to set this.

Fixes #269, fixes #250

This does essentially the same as #285 albeit keeping Apollos naming convention.

### DIFF
--- a/src/GraphQL.Client.Abstractions.Websocket/GraphQLWebSocketRequest.cs
+++ b/src/GraphQL.Client.Abstractions.Websocket/GraphQLWebSocketRequest.cs
@@ -14,7 +14,7 @@ namespace GraphQL.Client.Abstractions.Websocket
         public const string PAYLOAD_KEY = "payload";
 
         /// <summary>
-        /// The Identifier of the Response
+        /// The Identifier of the request
         /// </summary>
         public string Id
         {
@@ -34,7 +34,7 @@ namespace GraphQL.Client.Abstractions.Websocket
         /// <summary>
         /// The payload of the websocket request
         /// </summary>
-        public object Payload
+        public object? Payload
         {
             get => TryGetValue(PAYLOAD_KEY, out object value) ? value : null;
             set => this[PAYLOAD_KEY] = value;
--- a/src/GraphQL.Client.Abstractions.Websocket/GraphQLWebSocketRequest.cs
+++ b/src/GraphQL.Client.Abstractions.Websocket/GraphQLWebSocketRequest.cs
@@ -34,9 +34,9 @@ namespace GraphQL.Client.Abstractions.Websocket
         /// <summary>
         /// The payload of the websocket request
         /// </summary>
-        public GraphQLRequest Payload
+        public object Payload
         {
-            get => ContainsKey(PAYLOAD_KEY) ? (GraphQLRequest)this[PAYLOAD_KEY] : null;
+            get => ContainsKey(PAYLOAD_KEY) ? this[PAYLOAD_KEY] : null;
             set => this[PAYLOAD_KEY] = value;
         }
 
--- a/src/GraphQL.Client.Abstractions.Websocket/GraphQLWebSocketRequest.cs
+++ b/src/GraphQL.Client.Abstractions.Websocket/GraphQLWebSocketRequest.cs
@@ -18,7 +18,7 @@ namespace GraphQL.Client.Abstractions.Websocket
         /// </summary>
         public string Id
         {
-            get => ContainsKey(ID_KEY) ? (string)this[ID_KEY] : null;
+            get => TryGetValue(ID_KEY, out object value) ? (string)value : null;
             set => this[ID_KEY] = value;
         }
 
@@ -27,7 +27,7 @@ namespace GraphQL.Client.Abstractions.Websocket
         /// </summary>
         public string Type
         {
-            get => ContainsKey(TYPE_KEY) ? (string)this[TYPE_KEY] : null;
+            get => TryGetValue(TYPE_KEY, out object value) ? (string)value : null;
             set => this[TYPE_KEY] = value;
         }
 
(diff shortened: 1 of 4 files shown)


src/GraphQL.Client.Abstractions.Websocket/GraphQLWebSocketRequest.cs
```diff
@@ -34,9 +34,9 @@ public string Type
         /// <summary>
         /// The payload of the websocket request
         /// </summary>
-        public GraphQLRequest Payload
+        public object Payload
         {
-            get => ContainsKey(PAYLOAD_KEY) ? (GraphQLRequest)this[PAYLOAD_KEY] : null;
+            get => ContainsKey(PAYLOAD_KEY) ? this[PAYLOAD_KEY] : null;
```
sungam3r, line 39: I would prefer to change all usages of this `if (contains) then get()` pattern to `return TryGetValue(key, out v) ? v : null` to reduce dictionary key lookups.

sungam3r: Also editorial change on line 17:
> The Identifier of the Response

request, not response

src/GraphQL.Client/GraphQLHttpClientOptions.cs
```diff
@@ -56,5 +56,11 @@ public class GraphQLHttpClientOptions
         /// Configure additional websocket options (i.e. headers). This will not be invoked on Windows 7 when targeting .NET Framework 4.x. 
         /// </summary>
         public Action<ClientWebSocketOptions> ConfigureWebsocketOptions { get; set; } = options => { };
+
+        /// <summary>
+        /// The `ConnectionParams` object sent with the GQL_CONNECTION_INIT message on establishing a websocket connection.
+        /// See https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_connection_init.
+        /// </summary>
+        public object? WebSocketConnectionParams { get; set; } = null;
```
sungam3r, line 64: I advise you to change the type of this property to delegate. Often GQL_CONNECTION_INIT message is used to convey authentication information. This data is different for different users. Note that the properties above have already been made by delegates for the same reason - this is more flexible.

sungam3r: Also consider to converge naming. I can see `WebSocketConnectionParams`, `ConnectionParams`, `Payload` terms now both describing the same thing. I would prefer to use `Payload` term.

rose-a: I took the name `ConnectionParams` from the Apollo spec, to stay close to the naming used there... most "tutorials" are for Appollo anyway...

I think its confusing if we call this property on the options class just `Payload`...

rose-a: I renamed it to `SetWebSocketConnectionInitPayload`, I suppose that's what you meant in the first place...

src/GraphQL.Client/GraphQLHttpClientOptions.cs
```diff
@@ -56,5 +56,11 @@ public class GraphQLHttpClientOptions
         /// Configure additional websocket options (i.e. headers). This will not be invoked on Windows 7 when targeting .NET Framework 4.x. 
         /// </summary>
         public Action<ClientWebSocketOptions> ConfigureWebsocketOptions { get; set; } = options => { };
+
+        /// <summary>
+        /// Sets the `ConnectionParams` object sent with the GQL_CONNECTION_INIT message on establishing a GraphQL websocket connection.
+        /// See https://github.com/apollographql/subscriptions-transport-ws/blob/master/PROTOCOL.md#gql_connection_init.
+        /// </summary>
+        public Func<object?> SetWebSocketConnectionInitPayload { get; set; } = () => null;
```
sungam3r, line 64: 1. ConfigureWebSocketConnectionInitPayload ?
2. Maybe to pass some context info (at least the id of the request) into delegate - `Func<SomeContext, object?>ConfigureWebSocketConnectionInitPayload {get;set;} = _ => null; ` A separate class will allow you to expand the context in the future while maintaining backward compatibility with the delegate signature.

rose-a: We won't have the request-id in the future (the current implementation of the handshake is flawed, see #293). After #293 is fixed this will only be sent once on connecting the websocket, not for each subscription...

I think this way it's pretty similar to the javascript implementation...

sungam3r: I khow that this should be sent only once. Additional context won't hurt I think.

rose-a: But what context? I suspect you don't even need an id for the `GQL_CONNECTION_INIT` request... Everything else is known from the options class..

sungam3r: Then at least these options should be passed.

sungam3r (approved):

PR #295, opened 2020-10-08
Fix graphql websocket protocol

### LINKED ISSUE #293: Fix implementation details of the GraphQL over WebSocket Protocol

rose-a:
Fixes #293

### DIFF
--- a/src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
+++ b/src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
@@ -8,6 +8,7 @@ using System.Reactive.Disposables;
 using System.Reactive.Linq;
 using System.Reactive.Subjects;
 using System.Reactive.Threading.Tasks;
+using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
 using GraphQL.Client.Abstractions.Websocket;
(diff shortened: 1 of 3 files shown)


src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
```diff
@@ -468,9 +456,38 @@ private async Task ConnectAsync(CancellationToken token)
                 Debug.WriteLine($"new incoming message stream {_incomingMessages.GetHashCode()} created");
 
                 _incomingMessagesConnection = new CompositeDisposable(maintenanceSubscription, connection);
+                
+                var initRequest = new GraphQLWebSocketRequest
+                {
+                    Type = GraphQLWebSocketMessageType.GQL_CONNECTION_INIT,
+                    Payload = new GraphQLRequest()
```
sungam3r, line 463: new GraphQLRequest() ? It seems like there could be any type of payload.

rose-a: yeah, that should be replaced by the payload from #294... I'll fix this as soon as the former gets merged...

rose-a: Just didn't want to mix them up

sungam3r: ok

Issue #332, opened 2021-03-29
Problem: SendQueryAsync no response


masouri:
### Problem
As the title suggestest i send a graphql request, i do this in 3 different situations but in one of those 
the method SendQueryAsync just stops without any errors(tried try catch and breakpoints).

####Details
I have one Class to do this request, i use it mutliple times, but in one case it doesn't work and i have no idea
why, or how to debbug it 

###This is my Code

```C#
using GermanFishing.Model.GraphQLImages;
using GraphQL.Client.Http;
using GraphQL.Client.Serializer.Newtonsoft;
using System;
using System.Collections.Generic;
using System.IO;
using System.Text;
using System.Threading.Tasks;
using Xamarin.Forms;

namespace GermanFishing.ViewModel
{
    public  class UserViewModel
    {

        public User user_profile { get; set; }

        public UserViewModel()
        {

            user_profile = new User();

        }


        //Consume User Data from Graphql
        public async Task GetProfileData()
        {
            var client = new GraphQLHttpClient(new GraphQLHttpClientOptions
            {
                EndPoint = new Uri("https://www.german-fishing.de/graphql"),

            }, new NewtonsoftJsonSerializer());


            client.HttpClient.DefaultRequestHeaders.Add("Authorization", "Bearer " + Convert.ToString(Application.Current.Properties["Login"]));

            var request = new GraphQLHttpRequest
            {
                Query = @"
                           query {
                                  me {
                                    name
                                    unreadNotifications {
                                      data
                                    }
                                    abonnements {
                                      id
                                      place {
                                        id
                                        name
                                      }
                                    }
                                  }
                                }"
            };

            var response = await client.SendQueryAsync<UserImage>(request).ConfigureAwait(false);

            this.user_profile = response.Data.me;
            


        }

    }
}

```
Any suggestions how to approach this ?

sungam3r:
Can you turn on / check logs on the server side?

masouri:
> Can you turn on / check logs on the server side?

This is where it gets wacky, server gets a post request returns a 200 code,  nothing weird or irregular happening there

masouri:
###Additional Information
it could be that im in an Async Deadlock 
here is the Code where i acctually call it

```C#
using GermanFishing.ViewModel;
using System;
using System.Collections.Generic;
using System.Linq;
using System.Text;
using System.Threading.Tasks;

using Xamarin.Forms;
using Xamarin.Forms.Xaml;

[...]

PR #337, opened 2021-04-10
Fixed deadlock when calling SendQueryAsync synchronously #332

ds1709:
The following changes fixes deadlock problem from issue #332.

- Added `ConfigureAwait(false)` on `SendHttpRequestAsync` and `HttpClient.SendAsync`.

### DIFF
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -87,7 +87,8 @@ namespace GraphQL.Client.Http
                 Options.EndPoint.HasWebSocketScheme())
                 return await GraphQlHttpWebSocket.SendRequest<TResponse>(request, cancellationToken);
 
-            return await SendHttpRequestAsync<TResponse>(request, cancellationToken);
+            return await SendHttpRequestAsync<TResponse>(request, cancellationToken)
+                .ConfigureAwait(false);
         }
 
         /// <inheritdoc />
@@ -143,7 +144,8 @@ namespace GraphQL.Client.Http
             var preprocessedRequest = await Options.PreprocessRequest(request, this);
 
             using var httpRequestMessage = preprocessedRequest.ToHttpRequestMessage(Options, JsonSerializer);
-            using var httpResponseMessage = await HttpClient.SendAsync(httpRequestMessage, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+            using var httpResponseMessage = await HttpClient.SendAsync(httpRequestMessage, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
+                .ConfigureAwait(false);
 
             var contentStream = await httpResponseMessage.Content.ReadAsStreamAsync();
 


src/GraphQL.Client/GraphQLHttpClient.cs
```diff
@@ -87,7 +87,8 @@ public async Task<GraphQLResponse<TResponse>> SendQueryAsync<TResponse>(GraphQLR
                 Options.EndPoint.HasWebSocketScheme())
                 return await GraphQlHttpWebSocket.SendRequest<TResponse>(request, cancellationToken);
 
-            return await SendHttpRequestAsync<TResponse>(request, cancellationToken);
+            return await SendHttpRequestAsync<TResponse>(request, cancellationToken)
+                .ConfigureAwait(false);
```
sungam3r, line 91: Why only there and not two lines above also ?

src/GraphQL.Client/GraphQLHttpClient.cs
```diff
@@ -143,7 +144,8 @@ private async Task<GraphQLHttpResponse<TResponse>> SendHttpRequestAsync<TRespons
             var preprocessedRequest = await Options.PreprocessRequest(request, this);
 
             using var httpRequestMessage = preprocessedRequest.ToHttpRequestMessage(Options, JsonSerializer);
-            using var httpResponseMessage = await HttpClient.SendAsync(httpRequestMessage, HttpCompletionOption.ResponseHeadersRead, cancellationToken);
+            using var httpResponseMessage = await HttpClient.SendAsync(httpRequestMessage, HttpCompletionOption.ResponseHeadersRead, cancellationToken)
```
sungam3r, line 147: Why only there and not two lines below also ?

ds1709: Deadlock happens exactly because of `HttpClient.SendAsync` has no `ConfigureAwait(false)`. To resolve issue #332 it's enougth to call `ConfigureAwait(false)` just here. But idealy whole project should be reviewed and each `await` should be configured. First, it will prevent from potential deadlocks, and second, it may have positive impact on productivity.

sungam3r: > Deadlock happens exactly because of `HttpClient.SendAsync` has no `ConfigureAwait(false)`

Could you please write a test to verify this assertion ?

ds1709: Do you meen a unit test or just simple test application to be shure?

sungam3r: Unit test to be able to run on CI.

sungam3r: Yiu may call it unit or integration, whatever. The key part is to verify behavior on CI.

ds1709: I'm not sure there's somthing to test. Also I'm not sure it's posible to test `ConfigureAwait` or dedlock prevention. So I can't write it.

PR #343, opened 2021-04-26
Add PolymorphicTypenameConverter class to handle __typename deserialization 

lanwin:
Attached a helper class I found very useful to handle fragments and  __typename deserialization:

Usage:

```csharp
            [JsonConverter(typeof(DiscriminateEvent))]
            public Event Event { get; set; }

            class DiscriminateEvent : PolymorphicTypenameConverter<Event>
            {
                protected override Type Descriminator(string typename) =>
                    typename switch
                    {
                        "EventA" => typeof(EventA),
                        "EventB" => typeof(EventB),
                        _ => typeof(Event)
                    };
            }
```

I think there should be also a example within the readme cause its not obvious how to use it. 

But for now I want to know what you think first and if its a nice addition.

### DIFF
--- a/src/GraphQL.Client.Serializer.SystemTextJson/PolymorphicTypenameConverter.cs
+++ b/src/GraphQL.Client.Serializer.SystemTextJson/PolymorphicTypenameConverter.cs
@@ -1,0 +1,32 @@
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace GraphQL.Client.Serializer.SystemTextJson
+{
+    public abstract class PolymorphicTypenameConverter<T> : JsonConverter<T>
+    {
+        protected abstract Type Descriminator(string typename);
+
+        public override bool CanConvert(Type typeToConvert) =>
+            typeof(T).IsAssignableFrom(typeToConvert);
+
+        public override T Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            var clone = reader; // cause its a struct
+
+            if(clone.TokenType == JsonTokenType.StartObject)
+                clone.Read();
+            if(clone.TokenType != JsonTokenType.PropertyName || clone.GetString() != "__typename")
+                throw new JsonException();
+
+            clone.Read();
+            var type = Descriminator(clone.GetString());
+            object deserialize = JsonSerializer.Deserialize(ref reader, type, options);
+            return (T)deserialize;
+        }
+
+        public override void Write(Utf8JsonWriter writer, T obj, JsonSerializerOptions options) =>
+            throw new NotSupportedException();
+    }
+}


src/GraphQL.Client.Serializer.SystemTextJson/PolymorphicTypenameConverter.cs
```diff
@@ -0,0 +1,32 @@
...
+    public abstract class PolymorphicTypenameConverter<T> : JsonConverter<T>
+    {
+        protected abstract Type Descriminator(string typename);
+
+        public override bool CanConvert(Type typeToConvert) =>
+            typeof(T).IsAssignableFrom(typeToConvert);
+
+        public override T Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options)
+        {
+            var clone = reader; // cause its a struct
```
sungam3r, line 16: ?

src/GraphQL.Client.Serializer.SystemTextJson/PolymorphicTypenameConverter.cs
```diff
@@ -0,0 +1,32 @@
...
+
+            if(clone.TokenType == JsonTokenType.StartObject)
+                clone.Read();
+            if(clone.TokenType != JsonTokenType.PropertyName || clone.GetString() != "__typename")
```
sungam3r, line 20:
```suggestion
            if (clone.TokenType == JsonTokenType.StartObject)
                clone.Read();
            if (clone.TokenType != JsonTokenType.PropertyName || clone.GetString() != "__typename")

```

sungam3r, line 20: So it works only if I query `__typename` meta-field.

lanwin: Right. its designed to help with using Graphql Fragments. No other purpose.

src/GraphQL.Client.Serializer.SystemTextJson/PolymorphicTypenameConverter.cs
sungam3r: Ok. Better to put a comment into sources.

sungam3r: By itself, this class without descendants carries little benefit. You can make at least one descendant, which discovers all the classes in the declaring assembly.

src/GraphQL.Client.Serializer.SystemTextJson/PolymorphicTypenameConverter.cs
```diff
@@ -0,0 +1,33 @@
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace GraphQL.Client.Serializer.SystemTextJson
+{
+    public abstract class PolymorphicTypenameConverter<T> : JsonConverter<T>
```
rose-a, line 7: Perhaps we can find a class name which makes the use case for this converter a little bit more obvious (in that it's tied to the builtin GraphQL `__typename` field).

rose-a (changes requested):
Hi @lanwin, thanks for your contribution!

I think this can be very handy if you need to work with GraphQL interfaces and union types. 

We'd need a similar converter for *Newtonsoft.Json*, too.

src/GraphQL.Client.Serializer.SystemTextJson/PolymorphicTypenameConverter.cs
sungam3r: `FromTypenameConverter` ?

rose-a: `FromGraphQLTypenameConverter`? 😉

sungam3r:
@lanwin Would you like to continue working on it?

lanwin:
> @lanwin Would you like to continue working on it?

Yes but I am not sure yet whats expected from here.

sungam3r:
@rose-a ?

StevenDewey:
It would be great to see this or something like this supported! It is a great idea, though it looks like it has lost some momentum over the past year...

Keyslam:
Works like a charm when using [unions for error handling](https://sachee.medium.com/200-ok-error-handling-in-graphql-7ec869aec9bc). Would like to see it merged into the library. Cheers!

Issue #354, opened 2021-06-26
Testable Subscriptions with compatibility for TestServer


andreyleskov:
Hello everybody, 

I'm using graphql-dotnet project a lot and would like to thank you for the amazing product you build for the whole .Net community. 
My company is building GraphQl-first services and relies a lot on integration testing with TestSever. 

graphql-client Queries and Mutations work fine with TestServer because it is possible to pass an IHttpClient instance into GraphQlHttpClient constructor. 
But for subscription, there is no way to pass WebSocket from TestServer due to the not-configurable creation of ClientWebSocket at 
https://github.com/graphql-dotnet/graphql-client/blob/b366e22af112445452440c2c6c1d9978ecef97cb/src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs#L403

My suggestion is to add a new constructor  
```
    public GraphQLHttpWebSocket(Uri webSocketUri, GraphQLHttpClient client, Func<WebSocket> webSocketFactory)
    
```
to get better control over WebSocket creation. 
It would also allow moving condition compilation logic '#if NETFRAMEWORK'out of the GraphQLHttpWebSocket class later on, making it cleaner.

I see a lot of value in supporting subscriptions for TestServer. Our particular scenario is to issue asynchronous mutations and wait for some delayed effect via subscription. 

What do you think about such a suggestion? 
If you find this idea worthy, I can prepare a PR for your review and further discussion.

rose-a:
Hi, 

I agree the current setup for integration-testing against a webserver setup is not elegant.
Currently it's spinning up an actual webserver on a free port on the testserver. 

I'd love to see an elegant solution to this. Please go ahead and prove your concept in the integration tests project.

andreyleskov:
Hi rose-a, 

I'm glad you like the idea, will prepare the PR for your review this week.

rose-a:
I'm actually hoping for improved stability on the integration test through this, too (see #161). 😉

andreyleskov:
I already spotted this issue locally, not sure about its nature yet

andreyleskov:
Submitted https://github.com/graphql-dotnet/graphql-client/pull/357

rose-a:
Thanks for your work, I'll try to review this over the weekend...

andreyleskov:
@rose-a should provide more explanations or details to help you out a bit?

rose-a:
[...]

Issue #161, opened 2020-02-12
Unstable Tests


rose-a:
- [`GraphQL.Integration.Tests.WebsocketTest.CanReconnectWithSameObservable`](https://github.com/graphql-dotnet/graphql-client/blob/f5a5b1e4a3b88ca72ac16e1ed982b8107eb9b1f5/tests/GraphQL.Integration.Tests/WebsocketTest.cs#L112)
- [`GraphQL.Integration.Tests.WebsocketTest.CanConnectTwoSubscriptionsSimultaneously`](https://github.com/graphql-dotnet/graphql-client/blob/f5a5b1e4a3b88ca72ac16e1ed982b8107eb9b1f5/tests/GraphQL.Integration.Tests/WebsocketTest.cs#L182)

rose-a:
```
Test run for /home/runner/work/graphql-client/graphql-client/tests/GraphQL.Integration.Tests/bin/Release/netcoreapp3.1/GraphQL.Integration.Tests.dll(.NETCoreApp,Version=v3.1)
2020-03-04T14:35:18.8409254Z Microsoft (R) Test Execution Command Line Tool Version 16.3.0
2020-03-04T14:35:18.8412611Z Copyright (c) Microsoft Corporation.  All rights reserved.
2020-03-04T14:35:18.8430935Z 
2020-03-04T14:35:18.9592919Z Starting test execution, please wait...
2020-03-04T14:35:19.0052826Z 
2020-03-04T14:35:19.0053788Z A total of 1 test files matched the specified pattern.
2020-03-04T14:35:26.5724924Z [xUnit.net 00:00:06.22]     GraphQL.Integration.Tests.QueryAndMutationTests.Newtonsoft.PostRequestCanBeCancelled [FAIL]
2020-03-04T14:35:27.1115946Z [xUnit.net 00:00:06.76]     GraphQL.Integration.Tests.QueryAndMutationTests.SystemTextJson.PostRequestCanBeCancelled [FAIL]
2020-03-04T14:35:27.1433668Z   X GraphQL.Integration.Tests.QueryAndMutationTests.Newtonsoft.PostRequestCanBeCancelled [649ms]
2020-03-04T14:35:27.1435163Z   Error Message:
2020-03-04T14:35:27.1435904Z    Execution of the action should be less than 0.500s, but it required more than 0.500s and 320.7Âµs.
2020-03-04T14:35:27.1436003Z   Stack Trace:
[...]

PR #357, opened 2021-06-29
Add TestServer support for GraphQLHttpClient #354

andreyleskov:
This PR makes GraphQLHttpClient usable with TestServer.  

Usage: 
```
  TestServer testServer; //needs initiazation
  GraphQLHttpClientOptions options; //needs initiazation
  IGraphQLWebsocketJsonSerializer serializer;  //needs initiazation

  var graphQLHttpClient = testServer.CreateGraphQLHttpClient(options,serializer);
```

Implementation: 
 GraphQLHttpClient and GraphQLHttpWebSocket have a new internal constructor with a new factory parameter for connected WebSocket creation. 

```
Func<Uri, CancellationToken,Task<WebSocket>> connectedWebSocketFactory
```
 This parameter is used to provide TestWebSocket instance from TestServer. 
 There is a new project GraphQL.Client.TestHost with the extensions method to configure this factory for a given TestServer.

Discussion items: 

1. Need to check if net461 build is ok, as I checked it only with Mono under MacOS, would be nice to check with full .Net. 
2. There is `CanHandleConnectionTimeout` integration test that I could not make work for TestServer.  Seems like TestServer does not drop WebSocket connection on dispose. I could be wrong. Any help or suggestions are welcome. As a shortcut, I moved this test out of scope for TestServer
3. Way to inject WebSocket into GraphQLHttpClient. Now it is dictated by TestServer's WebSocketClient, and, maybe, there is a better way other than  GraphQLHttpClient internal constructor requiring complex factory setup and separate project usage. 
4. Tests stability, as mentioned in #161. I can see the same behavior for TestServer and think it is caused by synchronization inside tests themselves. It is not targeted here, but we can check it by running a dedicated server for each test.

### DIFF
--- a/src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
+++ b/src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
@@ -39,11 +39,7 @@ namespace GraphQL.Client.Http.Websocket
         private Task _initializeWebSocketTask = Task.CompletedTask;
         private readonly object _initializeLock = new object();
 
-#if NETFRAMEWORK
-		private WebSocket _clientWebSocket = null;
-#else
-        private ClientWebSocket _clientWebSocket = null;
-#endif
+ 		private WebSocket _clientWebSocket = null;
 
         #endregion
 
@@ -83,6 +79,25 @@ namespace GraphQL.Client.Http.Websocket
                 .Select(request => Observable.FromAsync(() => SendWebSocketRequestAsync(request)))
                 .Concat()
                 .Subscribe();
+
+            _connectedWebSocketFactory = async (uri, token) =>
+            {
+#if NETFRAMEWORK
+                var socket =  InitializeNetClientWebSocket();
+#else
+                var socket = InitializeNetCoreClientWebSocket();
+#endif
+                Debug.WriteLine($"opening websocket {socket.GetHashCode()} (thread {Thread.CurrentThread.ManagedThreadId})");
+                await socket.ConnectAsync(uri, token);
+                return socket;
+
+            };
+
+        }
+        internal GraphQLHttpWebSocket(Uri webSocketUri, GraphQLHttpClient client, Func<Uri, CancellationToken,Task<WebSocket>>? connectedWebSocketFactory):this(webSocketUri, client)
+        {
+            if(connectedWebSocketFactory!=null)
+               _connectedWebSocketFactory = connectedWebSocketFactory;
         }
 
         #region Send requests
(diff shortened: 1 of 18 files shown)


rose-a (commented):
Hi, just wanted to let you know that im currently looking into your MR...

The results from your new tests and those from the legacy ones are the same (which is good), but they are also noteably faster. I suspect the errors mentioned in #161 are the result of a threading issue with the test classes for the `IObservable`s, since the trigger for the observable happens in the same thread in which the expected result is awaited, which is a totally untypical use of `System.Reactive`. They're also sporadic (it's not always the same test that fails), which makes it even harder to find them.
I'm having the same issues in other projects im working on where I'm trying to test observables in a similar way.

I think the injection of a websocket factory might actually be a nice improvement, since it would allow us to move the dependency on `System.Net.WebSockets.Client.Managed` and all that conditional `#if NETFRAMEWORK` stuff to a separate package which is only ever needed when targeting Windows 7.

rose-a:
@andreyleskov please rebase your branch on release-v4

src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
```diff
@@ -442,8 +387,7 @@ private async Task ConnectAsync(CancellationToken token)
             {
                 await BackOff();
                 _stateSubject.OnNext(GraphQLWebsocketConnectionState.Connecting);
-                Debug.WriteLine($"opening websocket {_clientWebSocket.GetHashCode()} (thread {Thread.CurrentThread.ManagedThreadId})");
-                await _clientWebSocket.ConnectAsync(_webSocketUri, token);
+                _clientWebSocket = await _webSocketFactory.ConnectAsync(_webSocketUri, token);
```
sungam3r, line 390: @rose-a @andreyleskov regarding 4. Tests stability:

I recommend you **ALWAYS** use `ConfigureAwait(false)` in library code especially when using XUnit package as a test framework because of [`Xunit.Sdk.MaxConcurrencySyncContext`](https://github.com/xunit/xunit/blob/main/src/xunit.v3.core/Sdk/MaxConcurrencySyncContext.cs). In my case `maximumConcurrencyLevel` was 6 and some tests randomly failed. So please look through entire codebase and add `ConfigureAwait(false)` to awaitable calls. Perhaps it will help.

tests/GraphQL.Integration.Tests/WebsocketTests/BaseWithTimeout.cs
```diff
@@ -0,0 +1,99 @@
...
+using GraphQL.Client.Tests.Common.Chat;
+using GraphQL.Client.Tests.Common.FluentAssertions.Reactive;
+using GraphQL.Client.Tests.Common.Helpers;
+using GraphQL.Integration.Tests.Helpers;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace GraphQL.Integration.Tests.WebsocketTests
+{
+    public abstract class BaseWithTimeout:Base
```
sungam3r, line 19:
```suggestion
    public abstract class BaseWithTimeout : Base
```

src/GraphQL.Client/Websocket/GraphQLHttpWebSocket.cs
andreyleskov: Thanks a lot for the comment, it makes sense from my point of view. This PR is following the existing await style and focuses on just WebSockets. I assume repository owners prefer to do repository-wide changes as await style in a dedicated PR.

rose-a: im on to this in in #370...

andreyleskov:
Hey guys, can I do smth to make this PR move? Please let me know

jerry2007:
Hello,  what about this PR. I would like to use this...

alexander-jesner-AP:
Hi, @rose-a, is it possible to merge this PR? If not, how can we help to push it towards being mergeable?

tomachristian:
This would have been a lifesaver...

PR #369, opened 2021-10-29
Fix for double instantiation of GraphQLLocalExecutionClient

sungam3r:


rose-a:
> @rose-a Tests are unstable. I managed to pass tests only for second attempt. Also I've noticed that github CI creates nuget packages twice - on the build and test steps.

Yep, see #161, this seems to be some kind of threading issue with the way I'm trying to test the `IObservable` stuff, which also seems to depend on available system ressources (changes of passing are higher on a beefier machine, running the tests in question solitary on a development machine has almost 100% chance of passing... 😟

PR #370, opened 2021-11-01
Fix configure await

rose-a:
Related to #161, #337, #332

* Use `.ConfigureAwait(false)` on all awaits in library code
* Use updated [FluentAssertions.Reactive](https://www.nuget.org/packages/FluentAssertions.Reactive) for testing observables
* Remove "caching" of GraphQL subscriptions to fix test case `CanReconnectWithSameObservable`

Those overall changes seem to have considerably stabilized the unit tests, at least they run consistently across my dev machine and GitHub actions. 

@sungam3r I'd like a second opinion on the changes regarding the `CanReconnectWithSameObservable` test case, I'll try to elaborate a bit on how the behaviour of the GraphQL client has changed through this (relevant changes in cff7f5f9ec5932a823bff0fcc531048dcb314f88):

The test uses the *Chat* schema adopted from the server project (with some modifications), which uses a `ReplaySubject<1>` for it's [message stream](https://github.com/graphql-dotnet/graphql-client/blob/cff7f5f9ec5932a823bff0fcc531048dcb314f88/tests/GraphQL.Client.Tests.Common/Chat/Schema/IChat.cs#L27). The intentional design of this is that on subscribing to this stream the last message is played back on the client.

Up to now this library has created a *hot* observable for each subscription, cached it inside `GraphQLHttpClient` and returned the already created observable instance for repeated subscriptions with an equal request object. This caused the
aforementioned test case to fail on [this line](https://github.com/graphql-dotnet/graphql-client/blob/8cbe46ed4b2b155c4ce4b0eab199c8ef239fc771/tests/GraphQL.Integration.Tests/WebsocketTests/Base.cs#L246), where the second subscription expects to receive the replayed last chat message. But since the underlying GraphQL subscription did not terminate [here](https://github.com/graphql-dotnet/graphql-client/blob/8cbe46ed4b2b155c4ce4b0eab199c8ef239fc771/tests/GraphQL.Integration.Tests/WebsocketTests/Base.cs#L241) but instead was reused, the server did not even notice that the subscription was disposed on the client side (don't ask me how this test has passed sometimes in the past 🙈 ).

Now the client does not try to reuse subscriptions any more and passes `.Subscribe()` and `.Dispose()` transparenty to the server. This includes creating two identical simultaneous subscriptions, which would then send the same data twice over the websocket connection.

IMO the old behaviour introduced incomprehensible behaviour, which is much worse than sending data twice, which can and should be prevented by the client application code.

I'd appreciate your feedback.


sungam3r:
Sorry but I don't use subscriptions.

PR #417, opened 2022-07-16
Move IDisposable to implementation

sungam3r:
Normally the caller should not interact with `IDisposable` provided via `IGraphQLClient` interface.


rose-a (approved):
makes sense

PR #419, opened 2022-07-18
Allow non-200 HTTP status codes

sungam3r:
See https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#status-codes . Spec allow to return 400 i ncase of validation errors. GraphQL.NET / server support it.

### DIFF
--- a/src/GraphQL.Client/GraphQLHttpClientOptions.cs
+++ b/src/GraphQL.Client/GraphQLHttpClientOptions.cs
@@ -52,7 +52,7 @@ namespace GraphQL.Client.Http
         /// <summary>
         /// Delegate to determine if GraphQL response may be properly deserialized into <see cref="GraphQLResponse{T}"/>.
         /// </summary>
-        public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r => r.IsSuccessStatusCode;
+        public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r => r.IsSuccessStatusCode || r.StatusCode == HttpStatusCode.BadRequest;
 
         /// <summary>
         /// This callback is called after successfully establishing a websocket connection but before any regular request is made.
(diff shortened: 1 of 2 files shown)


sungam3r:
@rose-a Please review. This feature blocks me from migrating some of our services to GraphQL.NET v5 since all validation errors return 400 instead of 200 now.

src/GraphQL.Client/GraphQLHttpClientOptions.cs
```diff
@@ -50,12 +50,17 @@ public class GraphQLHttpClientOptions
             Task.FromResult(request is GraphQLHttpRequest graphQLHttpRequest ? graphQLHttpRequest : new GraphQLHttpRequest(request));
 
         /// <summary>
-        /// This callback is called after successfully establishing a websocket connection but before any regular request is made. 
+        /// Delegate to determine if GraphQL response may be properly deserialized into <see cref="GraphQLResponse{T}"/>.
+        /// </summary>
+        public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r => r.IsSuccessStatusCode;
```
rose-a, line 55: Shouldn't we accept status 400 by default? As far as I can see that's whats returned when query validation fails in [graphql-dotnet/server](https://github.com/graphql-dotnet/server/blob/83c8ccaf12d259e2f3b65c9ce1f2982e95eb5e3b/src/Transports.AspNetCore/GraphQLHttpMiddleware.cs#L281)...

I think we should aim for default compatibility between those two repos...

```suggestion
        public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r => r.IsSuccessStatusCode || r.StatusCode == HttpStatusCode.BadRequest;
```

rose-a (changes requested):

src/GraphQL.Client/GraphQLHttpClientOptions.cs
sungam3r: I'm fine to accept 400 codes but it is some sort of semi-breaking changes and suites more if published in v5 some time ago.

sungam3r:
@rose-a If you fine with proposed changes I would like to publish fresh packages ASAP.

sungam3r:
👍

PR #429, opened 2022-07-26
Fix default behavior for 401/403 http response codes

sungam3r:
This PR aligns behavior with server project that returns 401/403.

### DIFF
--- a/src/GraphQL.Client/GraphQLHttpClientOptions.cs
+++ b/src/GraphQL.Client/GraphQLHttpClientOptions.cs
@@ -55,7 +55,9 @@ public class GraphQLHttpClientOptions
     /// Note that compatible to the draft graphql-over-http spec GraphQL Server MAY return 4xx status codes (401/403, etc.)
     /// with well-formed GraphQL response containing errors collection.
     /// </summary>
-    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r => r.IsSuccessStatusCode || r.StatusCode == HttpStatusCode.BadRequest || r.Content.Headers.ContentType.MediaType == "application/graphql+json";
+    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r =>
+        // Why not application/json? See https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#processing-the-response
+        r.IsSuccessStatusCode || r.StatusCode == HttpStatusCode.BadRequest || r.Content.Headers.ContentType.MediaType == "application/graphql+json";
 
     /// <summary>
     /// This callback is called after successfully establishing a websocket connection but before any regular request is made.
--- a/src/GraphQL.Client/GraphQLHttpClientOptions.cs
+++ b/src/GraphQL.Client/GraphQLHttpClientOptions.cs
@@ -52,8 +52,10 @@ public class GraphQLHttpClientOptions
 
     /// <summary>
     /// Delegate to determine if GraphQL response may be properly deserialized into <see cref="GraphQLResponse{T}"/>.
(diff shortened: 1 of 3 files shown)


src/GraphQL.Client/GraphQLHttpClientOptions.cs
```diff
@@ -52,8 +52,10 @@ public class GraphQLHttpClientOptions
 
     /// <summary>
     /// Delegate to determine if GraphQL response may be properly deserialized into <see cref="GraphQLResponse{T}"/>.
+    /// Note that compatible to the draft graphql-over-http spec GraphQL Server MAY return 4xx status codes (401/403, etc.)
+    /// with well-formed GraphQL response containing errors collection.
     /// </summary>
-    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r => r.IsSuccessStatusCode || r.StatusCode == HttpStatusCode.BadRequest;
+    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r => r.IsSuccessStatusCode || r.StatusCode == HttpStatusCode.BadRequest || r.Content.Headers.ContentType.MediaType == "application/graphql+json";
```
rose-a, line 58: This tries to parse all responses in which the server manages to set the content type header to `application/graphql+json` (including 5xx responses). Are u sure this is a "bulletproof" assumption?

I think we should at least catch parsing exceptions in [GraphQLHttpClient](https://github.com/graphql-dotnet/graphql-client/blob/758ede0e1773dafc9f9c662e495e89fe852a9786/src/GraphQL.Client/GraphQLHttpClient.cs#L128) and wrap those up in a [GraphQLHttpRequestException](https://github.com/graphql-dotnet/graphql-client/blob/89ba50d68e14c31910d4cf4e384dbe84135b0de8/src/GraphQL.Client/GraphQLHttpRequestException.cs) to pass the raw response content to the user (this probably won't hurt, anyway).

rose-a:
What's missing, too, is that this client doesn't set the *Accept* request header as defined in the [spec draft](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#accept), so the server will only respond with `application/graphql+json` if you specify the accept header in your application code.

We should set it to `application/graphql+json; charset=utf-8, application/json; charset=utf-8` as recommended. I propose to do that in [GraphQLHttpRequest.ToHttpRequestMessage](https://github.com/graphql-dotnet/graphql-client/blob/758ede0e1773dafc9f9c662e495e89fe852a9786/src/GraphQL.Client/GraphQLHttpRequest.cs#L35).

src/GraphQL.Client/GraphQLHttpClientOptions.cs
sungam3r: If content type is `application/graphql+json` then the response **can** be deserialized as a valid graphql response. 

Regarding 5xx error codes from server - see https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#applicationgraphqljson :
изображение

sungam3r:
OK, will do tomorrow.

src/GraphQL.Client/GraphQLHttpRequest.cs
```diff
@@ -38,6 +39,9 @@ public virtual HttpRequestMessage ToHttpRequestMessage(GraphQLHttpClientOptions
         {
             Content = new StringContent(serializer.SerializeToString(this), Encoding.UTF8, options.MediaType)
         };
+        message.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/graphql+json"));
```
sungam3r, line 42: `MediaTypeWithQualityHeaderValue` does not accept `application/graphql+json; charset=utf-8`. It throws `System.FormatException : The format of value 'application/graphql+json; charset=utf-8' is invalid`. I added Accept-Charset header that should work the same.

rose-a (approved):
looks good to me..

PR #493, opened 2022-11-25
Graphql transport ws

joao-avelino:
This pull request adds the ``graphql-transport-ws``  [sub-protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md) to the client.

We tested subscriptions with the Graphql Yoga server.

Sample Program.cs:

```Csharp
using GraphQL;
using GraphQL.Client.Http;
using GraphQL.Client.Serializer.Newtonsoft;
using GraphQL.Client.Http.Websocket;

IObservable<GraphQLResponse<dynamic>> _subscriptionStream = default!;
IDisposable _subscription  = default!;

var graphQLHttpClientOptions = new GraphQLHttpClientOptions
{
    EndPoint = new Uri("https://localhost:8080/graphql"),
    WebSocketEndPoint = new Uri("wss://localhost:8080/mesh/graphql"),
    WebSocketProtocol = WebSocketProtocols.GRAPHQL_TRANSPORT_WS_PROTOCOL //or WebSocketProtocols.DEPRECATED_GRAPHQL_WS_PROTOCOL for old "graphql-ws"
};

var httpClient = new HttpClient();

var graphQLClient = new GraphQLHttpClient(graphQLHttpClientOptions, new NewtonsoftJsonSerializer(), httpClient);

var request = new GraphQLRequest
{
    Query = @"
subscription TestSubscription{
  userAdded
}"
};

_subscriptionStream = graphQLClient.CreateSubscriptionStream<dynamic>(request);
_subscription = _subscriptionStream.Subscribe(response =>
{
    Console.WriteLine(response.Data);
});


await Task.Delay(-1);

````

This pull request was developed at [μRoboptics](https://www.roboptics.pt/) and [evismo AG](https://evismo.com/en_GB/).

### CODE REVIEW

**Review by rose-a — CHANGES_REQUESTED** [review]
Hi, thanks for submitting this PR, this is a long time outstanding feature this library needs! 👏 ❤️ 

About the code:

1. I think it would keep the code cleaner if you would split up the `GraphQlHttpWebSocket` into an abstract base class which contains all the common code and then a derived implementation class for each websocket protocol.

2. I would not name the `graphql-ws` protocol "deprecated" throughout the code, this might irritate users which need to use it (because the server only supports the old one). It's still widely used, and all the subscription transport specifications are only somewhat semi-official. I'd just call them by their name (and ommit the word deprecated, keep that in the comment)

**Comment by joao-avelino** [comment]
Hi @rose-a. Thanks for your feedback.

I'll follow your suggestions and commit them next Monday :)

**Comment by sungam3r** [comment]
Just for information - GraphQL.NET server supports both protocols.

**Comment by sungam3r** [comment]
> I think it would keep the code cleaner if you would split up the GraphQlHttpWebSocket into an abstract base class which contains all the common code and then a derived implementation class for each websocket protocol.

👍  For reference - https://github.com/graphql-dotnet/server/tree/master/src/Transports.AspNetCore/WebSockets  , see `BaseSubscriptionServer` and 2 derived classes.

**Comment by joao-avelino** [comment]
Hello @rose-a  and @sungam3r . I've just made the commit. The default behavior is to use the old "graphql-ws" to avoid breaking backwards compatibility (if no WebSocketProtocol is provided to GraphQLHttpClientOptions).

I only saw the reference today, @sungam3r (although I've seen some tests for graphql-transport-w before). If something is not working as intended, please let me know and I'll try to follow the server's implementation more closely.

**Comment by sungam3r** [comment]
I don't use subscriptions, @Shane32 does.

**Comment by Shane32** [comment]
I don't use the GraphQL-Client nuget package, however, I did build the current implementation of subscription support within GraphQL.NET Server 7.x and have a few comments, listed below.  Please note that I have not reviewed the design of this project, nor this PR, so please excuse me if some of these comments are out of place.

- The WebSockets protocol supports client negotiation.  The client should provide both `graphql-ws` and `graphql-transport-ws` protocols to the server, from which the server can select the best appropriate protocol.  The negotiated protocol is visible in 
the [WebSocket.SubProtocol](https://learn.microsoft.com/en-us/dotnet/api/system.net.websockets.websocket.subprotocol?view=net-7.0) property.  The user should not need to select an implementation.  GraphQL.NET Server 7.x will select the newer protocol while prior versions will select the older protocol.  It should also be compatible with any other GraphQL WebSocket server.  See https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism under the section Sec-WebSocket-Protocol.

- I found the most notable non-obvious design difference in the two protocols to be in the response of the `error` / `GQL_ERROR` message.  The newer protocol explicitly returns an list of GraphQL error objects, while I'm not sure what exactly the older protocol should return.  See the [newer protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md#error).  GraphQL.NET Server 7.x currently returns an entire GraphQL response object (same as `GQL_DATA`) for the older protocol; this may or may not be correct, but was based on GraphQL.NET Server 6.x behavior.  Assuming errors are deserialized, I suggest testing for each protocol to be sure the format is as expected.

- For my own needs, I found that an `AsyncEnumerable<T>` was easier me to consume a GraphQL subscription than an `IObservable<T>`.  It returns a stream of messages, can indicate a successful completion of the source stream (if applicable), and can indicate a problem via a typed exception.  It's also extremely easy to implement (assuming a dedicated HTTP connection for each subscription), with all straightforward `await` calls.  There's also no ambiguity as to if/when the connection disposes; proper teardown is guaranteed due to simple `using` calls.  Connecting two servers together becomes as simple as this:

```cs
public class MySyncService : BackgroundService
{
    private async Task ExecuteAsync(CancellationToken stoppingToken)
    {
        var request = new GraphQLRequest {
            Query = QUERY,
            OperationName = "opSubscription",
        };
        while (!stoppingToken.IsCancellationRequested) {
            try {
                await Task.Delay(5000, stoppingToken);
                await foreach (var data in _graphQlClient.SubscribeAsync<MyResponse>(request, stoppingToken).ConfigureAwait(false)) {
                    // do some stuff
                }
            } catch (OperationCanceledException) when (stoppingToken.IsCancellationRequested) {
                throw;
            } catch (Exception ex) {
                // log caught exceptions here; then wait 5 seconds and retry connection
            }
        }
    }
}
```

I'm sure `IObservable<T>`-based implementations are not so much different, but that's what worked for me.  My entire implementation is only a couple hundred lines of code.

Shane32:
> If something is not working as intended, please let me know and I'll try to follow the server's implementation more closely.

The implementation of the newer protocol within GraphQL.NET Server 7.x should be very precisely aligned with the [official protocol](https://github.com/enisdenjo/graphql-ws/blob/master/PROTOCOL.md).  Please let me know if you find any discrepancies.

FYI, you can see the main message loop here (for the newer protocol):

https://github.com/graphql-dotnet/server/blob/6cb41a5c1b7ca9fabaf1fce2300e8645b9f3d423/src/Transports.AspNetCore/WebSockets/GraphQLWs/SubscriptionServer.cs#L73

rose-a:
@joao-avelino The websocket integration tests against GraphQL.NET Server are here:

https://github.com/graphql-dotnet/graphql-client/blob/master/tests/GraphQL.Integration.Tests/WebsocketTests/Base.cs

I shouldn't be too hard to add another derived test class which uses `graphql-transport-ws` against the server lib, you'll have to add another [IntegrationServerTestFixture](https://github.com/graphql-dotnet/graphql-client/blob/master/tests/GraphQL.Integration.Tests/Helpers/IntegrationServerTestFixture.cs) subclass to configure it this way.

joao-avelino:
Hello,

Thanks @Shane32 for the insightful comments. I'll try to improve the code. And thanks @rose-a for the websocket integration test base classe. 

I'll try to find some time to improve the code following your points, however it may take some time since I have some pressing matters at work right now.

rose-a:
closing in favor of #539

PR #537, opened 2023-04-06
Add an interface for websocket specific transport methods.

ben-voss:
Following on from this discussion: https://github.com/graphql-dotnet/graphql-client/discussions/526 this adds a new interface for the WebSocket transport specific methods defined on GraphQLHttpClient.

The main reason for this is so users can abstract the implementation from the interfaces and substitute them with mocks for unit testing.  This also has a side effect of refining the separation between WebSocket and Http transports.

### DIFF
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -6,7 +6,7 @@ using GraphQL.Client.Http.Websocket;
 
 namespace GraphQL.Client.Http;
 
-public class GraphQLHttpClient : IGraphQLClient, IGraphQLWebSocketClient, IDisposable
+public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
 {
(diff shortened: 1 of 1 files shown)


src/GraphQL.Client/GraphQLHttpClient.cs
```diff
@@ -6,7 +6,7 @@
 
 namespace GraphQL.Client.Http;
 
-public class GraphQLHttpClient : IGraphQLClient, IDisposable
+public class GraphQLHttpClient : IGraphQLClient, IGraphQLWebSocketClient, IDisposable
```
rose-a, line 9:
```suggestion
public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
```

rose-a (changes requested):

ben-voss:
I've accepted the requested change.  Is there anything else I need to do for this PR to be merged?

PR #539, opened 2023-04-11
Support graphql-transport-ws websocket protocol

rose-a:
This PR picks up #493, rebases on current master, adds unit tests, resolves the errors and does some refactoring.

Not yet implemented:

- [x]  protocol auto-negotiation (needs a complete refactoring of this PRs changes which separates websocket and protocol handling to be able to select the protocol from within the GraphQLHttpWebSocket class)
- [x]  client-side ping

@Shane32 Regarding ping/pong whats your opinion on how the payload of a ping should be handled? Its not explicitly stated in the spec, but my current implementation reflects the payload of the ping request in the corresponding pong request. 
Since the spec explicitly allows sending pongs at any time, this allows the server to correlate pongs with sent pings to use this for metrics etc.

### DIFF
--- a/src/GraphQL.Client/Websocket/GraphQLWSProtocolHandler.cs
+++ b/src/GraphQL.Client/Websocket/GraphQLWSProtocolHandler.cs
@@ -8,7 +8,7 @@ using GraphQL.Client.Abstractions.Websocket;
 
 namespace GraphQL.Client.Http.Websocket;
 
-internal class GraphQLWSProtocolHandler: IWebsocketProtocolHandler
+internal class GraphQLWSProtocolHandler : IWebsocketProtocolHandler
 {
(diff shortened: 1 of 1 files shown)


Shane32:
> how the payload of a ping should be handled

> reflects the payload of the ping request

I think that's a good idea.  I could implement the same in GraphQL.NET Server 7, although it is not so now.

Shane32:
Just FYI, when server-side keep-alive is enabled within GraphQL.NET Server 7, only pongs are sent, and only when the specified amount of time has elapsed from the last sent or received message.

rose-a:
@sungam3r pls review for code sanity ;)
@Shane32 thanks for your input

rose-a:
Already implemented and enabled by default in v6.0.0

Issue #516, opened 2023-02-13
GraphQLHttpClient


alex-m-git:
Hello everyone!

I have a small question about `GraphQLHttpClient`.
Is `GraphQLHttpClient `thread safe?

In any examples that I found on the Internet, they always create a new instance for each request.
Perhaps this is not correct and I can reduce the overhead and not create a new instance.
For example, I could register `GraphQLHttpClient` as singleton in DI.

rose-a:
Hi, the intended use of `GraphQLHttpClient` is to keep one instance alive per endpoint (obvious in case you're operating full websocket, but also true for regular requests).

So I recommend you register it as singleton in DI.

alex-m-git:
@rose-a 
Thanks a lot for the fast reply :)

PR #555, opened 2023-04-24
Add APQ support

### LINKED ISSUE #548: Built-in support for APQ (Persisted Queries)
APQ (Automatic Persisted Queries) are strandard and also a requirement for some GraphQL servers and they are also implemented in the graphql-server since 5.3

https://github.com/graphql-dotnet/graphql-dotnet/pull/3134

It would be cool to be able to use this feature out of box with the .net client
Either building the query with some helper methods first, or with opt-in automatic conversion

https://www.apollographql.com/docs/apollo-server/performance/apq/

sungam3r:
fixes #548 

@rose-a @Shane32 please review, especially TODO note about extensions type

Also I would like to note that design of GraphQL client from this repo allows to send arbitrary queries to the server since caller may build arbitrary strings, for example, ones with concatenated literals with different values
```graphql
query { user(id: "1234567") { name } }
```

Server cache for APQ may be polluted. In my in-house GraphQL client I do not allow to pass arbitrary strings. Instead this client autogenerates query based on used arguments in autogenerated methods and places all values into variables so the query text itself is already constant value like
```graphql
query q($userId: String!) { user(id: $userId) { name } }
```

Nevertheless, I'm fine to add general possibility to use APQ here.

### DIFF
--- a/src/GraphQL.Client/GraphQL.Client.csproj
+++ b/src/GraphQL.Client/GraphQL.Client.csproj
@@ -22,10 +22,6 @@
     <PackageReference Include="System.Net.WebSockets.Client.Managed" Version="1.0.22" />
   </ItemGroup>
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="System.Memory" Version="4.5.5" />
-  </ItemGroup>
-
   <ItemGroup>
     <ProjectReference Include="..\GraphQL.Client.Abstractions.Websocket\GraphQL.Client.Abstractions.Websocket.csproj" />
     <ProjectReference Include="..\GraphQL.Client.Abstractions\GraphQL.Client.Abstractions.csproj" />
--- a/src/GraphQL.Client/GraphQL.Client.csproj
+++ b/src/GraphQL.Client/GraphQL.Client.csproj
@@ -22,6 +22,10 @@
     <PackageReference Include="System.Net.WebSockets.Client.Managed" Version="1.0.22" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.Memory" Version="4.5.5" />
+  </ItemGroup>
+
   <ItemGroup>
     <ProjectReference Include="..\GraphQL.Client.Abstractions.Websocket\GraphQL.Client.Abstractions.Websocket.csproj" />
     <ProjectReference Include="..\GraphQL.Client.Abstractions\GraphQL.Client.Abstractions.csproj" />
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -169,9 +169,7 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
                 // GraphQL server supports APQ!
 
                 // Alas, for the first time we did not guess and in vain removed Query, so we return Query and
-                // send request again. This is one-time "cache miss", not so scary. By the way, after a second call,
-                // there is a probability of receiving PERSISTED_QUERY_NOT_FOUND error again, but it is so ghostly
-                // that this can and should be neglect so not to complicate the code without the need.
+                // send request again. This is one-time "cache miss", not so scary.
                 request.Query = savedQuery;
                 return await SendHttpRequestAsync<TResponse>(request, cancellationToken);
             }
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -17,12 +17,6 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
     private readonly CancellationTokenSource _cancellationTokenSource = new();
 
     private readonly bool _disposeHttpClient = false;
-
-    /// <summary>
-    /// This flag is used to completely disable APQ when GraphQL server does not support it.
-    /// </summary>
-    private bool _apqDisabledPerSession;
-
     /// <summary>
     /// the json serializer
     /// </summary>
@@ -37,6 +31,12 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
     /// The Options	to be used
     /// </summary>
     public GraphQLHttpClientOptions Options { get; }
+
+    /// <summary>
+    /// This flag is set to <see langword="true"/> when an error has occurred on an APQ and <see cref="GraphQLHttpClientOptions.DisableAPQ"/>
+    /// has returned <see langword="true"/>. To reset this, the instance of <see cref="GraphQLHttpClient"/> has to be disposed and a new one must be created.
+    /// </summary>
+    public bool APQDisabledForSession { get; private set; }
 
     /// <inheritdoc />
     public IObservable<Exception> WebSocketReceiveErrors => GraphQlHttpWebSocket.ReceiveErrors;
@@ -99,7 +99,7 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
         string? savedQuery = request.Query;
         bool useAPQ = false;
 
-        if (request.Query != null && !_apqDisabledPerSession && Options.EnableAutomaticPersistedQueries(request))
+        if (request.Query != null && !APQDisabledForSession && Options.EnableAutomaticPersistedQueries(request))
         {
             // https://www.apollographql.com/docs/react/api/link/persisted-queries/
             useAPQ = true;
@@ -129,7 +129,7 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
             {
                 // GraphQL server either supports APQ of some other version, or does not support it at all.
                 // Send a request for the second time. This is better than returning an error. Let the client work with APQ disabled.
-                _apqDisabledPerSession = Options.DisableAPQ(response);
+                APQDisabledForSession = Options.DisableAPQ(response);
                 request.Query = savedQuery;
                 return await SendQueryInternalAsync<TResponse>(request, cancellationToken);
             }
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -19,6 +19,11 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
     private readonly bool _disposeHttpClient = false;
 
     /// <summary>
+    /// This flag is used to completely disable APQ when GraphQL server does not support it.
+    /// </summary>
+    private bool _useAPQ = true;
+
+    /// <summary>
     /// the json serializer
     /// </summary>
     public IGraphQLWebsocketJsonSerializer JsonSerializer { get; }
@@ -89,7 +94,7 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
     {
         return Options.UseWebSocketForQueriesAndMutations || Options.WebSocketEndPoint is not null && Options.EndPoint is null || Options.EndPoint.HasWebSocketScheme()
             ? await GraphQlHttpWebSocket.SendRequestAsync<TResponse>(request, cancellationToken).ConfigureAwait(false)
-            : await SendHttpRequestAsync<TResponse>(request, cancellationToken).ConfigureAwait(false);
+            : await SendAPQHttpRequestAsync<TResponse>(request, cancellationToken).ConfigureAwait(false);
     }
 
     /// <inheritdoc />
@@ -123,6 +128,62 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
     public Task SendPongAsync(object? payload) => GraphQlHttpWebSocket.SendPongAsync(payload);
 
     #region Private Methods
+
+    private async Task<GraphQLHttpResponse<TResponse>> SendAPQHttpRequestAsync<TResponse>(GraphQLRequest request, CancellationToken cancellationToken = default)
+    {
+        cancellationToken.ThrowIfCancellationRequested();
+
+        var savedQuery = request.Query;
+        bool useAPQ = false;
+
+        if (request.Query != null && _useAPQ && Options.EnableAutomaticPersistedQueries(request))
+        {
+            // https://www.apollographql.com/docs/react/api/link/persisted-queries/
+            const int APQ_SUPPORTED_VERSION = 1;
+            useAPQ = true;
+            // TODO: I suggest to change request.Extension type to public Dictionary<string, object?>?
+            //request.Extensions ??= new();
+            //request.Extensions["persistedQuery"] = new Dictionary<string, object>
+            //{
+            //    ["version"] = APQ_SUPPORTED_VERSION,
+            //    ["sha256Hash"] = Hash.Compute(request.Query),
+            //};
+            request.Extensions ??= new
+            {
+                version = APQ_SUPPORTED_VERSION,
+                sha256Hash = Hash.Compute(request.Query),
+            };
+            request.Query = null;
+        }
+
+        var response = await SendHttpRequestAsync<TResponse>(request, cancellationToken);
+
+        if (useAPQ && response.Errors?.Length > 0)
+        {
+            if (response.Errors.Any(error => string.Equals(error.Message, "PersistedQueryNotFound", StringComparison.CurrentCultureIgnoreCase)))
+            {
+                // GraphQL server supports APQ!
+
+                // Alas, for the first time we did not guess and in vain removed Query, so we return Query and
+                // send request again. This is one-time "cache miss", not so scary. By the way, after a second call,
+                // there is a probability of receiving PERSISTED_QUERY_NOT_FOUND error again, but it is so ghostly
+                // that this can and should be neglect so not to complicate the code without the need.
+                request.Query = savedQuery;
+                return await SendHttpRequestAsync<TResponse>(request, cancellationToken);
+            }
+            else if (response.Errors.Any(error => string.Equals(error.Message, "PersistedQueryNotSupported") ||
+                     response.Errors.Any(error => string.Equals(error.Message, "GraphQL query is missing.")))) // GraphQL.NET specific error message
+            {
+                // GraphQL server either supports APQ of some other version, or does not support it at all.
+                // Send a request for the second time. This is better than returning an error. Let the client work with APQ disabled.
+                _useAPQ = false;
+                request.Query = savedQuery;
+                return await SendHttpRequestAsync<TResponse>(request, cancellationToken);
+            }
+        }
+
+        return response;
+    }
 
     private async Task<GraphQLHttpResponse<TResponse>> SendHttpRequestAsync<TResponse>(GraphQLRequest request, CancellationToken cancellationToken = default)
     {
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -21,7 +21,7 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
     /// <summary>
     /// This flag is used to completely disable APQ when GraphQL server does not support it.
     /// </summary>
-    private bool _APQdisabledPerSession;
+    private bool _apqDisabledPerSession;
 
     /// <summary>
     /// the json serializer
@@ -136,7 +136,7 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
         var savedQuery = request.Query;
         bool useAPQ = false;
 
-        if (request.Query != null && !_APQdisabledPerSession && Options.EnableAutomaticPersistedQueries(request))
+        if (request.Query != null && !_apqDisabledPerSession && Options.EnableAutomaticPersistedQueries(request))
         {
             // https://www.apollographql.com/docs/react/api/link/persisted-queries/
             const int APQ_SUPPORTED_VERSION = 1;
@@ -167,7 +167,7 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
             {
                 // GraphQL server either supports APQ of some other version, or does not support it at all.
                 // Send a request for the second time. This is better than returning an error. Let the client work with APQ disabled.
-                _APQdisabledPerSession = Options.DisableAPQ(response);
+                _apqDisabledPerSession = Options.DisableAPQ(response);
                 request.Query = savedQuery;
                 return await SendHttpRequestAsync<TResponse>(request, cancellationToken);
             }
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -150,9 +150,13 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
             //};
             request.Extensions ??= new
             {
-                version = APQ_SUPPORTED_VERSION,
-                sha256Hash = Hash.Compute(request.Query),
+                persistedQuery = new
+                {
+                    version = APQ_SUPPORTED_VERSION,
+                    sha256Hash = Hash.Compute(request.Query),
+                }
             };
+
             request.Query = null;
         }
 
--- a/src/GraphQL.Client/GraphQLHttpClient.cs
+++ b/src/GraphQL.Client/GraphQLHttpClient.cs
@@ -96,19 +96,15 @@ public class GraphQLHttpClient : IGraphQLWebSocketClient, IDisposable
     {
         cancellationToken.ThrowIfCancellationRequested();
 
-        string? savedQuery = request.Query;
+        string? savedQuery = null;
         bool useAPQ = false;
 
         if (request.Query != null && !APQDisabledForSession && Options.EnableAutomaticPersistedQueries(request))
         {
             // https://www.apollographql.com/docs/react/api/link/persisted-queries/
             useAPQ = true;
-            request.Extensions ??= new();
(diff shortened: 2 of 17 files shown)


src/GraphQL.Client/GraphQLHttpClientOptions.cs
```diff
@@ -87,4 +87,11 @@ public class GraphQLHttpClientOptions
     /// </summary>
     public ProductInfoHeaderValue? DefaultUserAgentRequestHeader { get; set; }
         = new ProductInfoHeaderValue(typeof(GraphQLHttpClient).Assembly.GetName().Name, typeof(GraphQLHttpClient).Assembly.GetName().Version.ToString());
+
+    /// <summary>
+    /// Delegate permitting use of <see href="https://www.apollographql.com/docs/react/api/link/persisted-queries/">Automatic Persisted Queries (APQ)</see>.
+    /// By default returns <see langword="true" /> for all requests. Note that GraphQL server should support APQ. Otherwise, the client disables APQ completely
+    /// after an unsuccessful attempt to send an APQ request and then send only regular requests.
+    /// </summary>
+    public Func<GraphQLRequest, bool> EnableAutomaticPersistedQueries { get; set; } = _ => true;
```
Shane32, line 96: Don't know if I would personally default it to enabled

sungam3r: The idea here to use APQ if server is ready for it. If not - client fallbacks to regular call for all subsequent requests.

src/GraphQL.Client/GraphQLHttpClient.cs
```diff
@@ -111,6 +116,62 @@ public IObservable<GraphQLResponse<TResponse>> CreateSubscriptionStream<TRespons
...
+
+                // Alas, for the first time we did not guess and in vain removed Query, so we return Query and
+                // send request again. This is one-time "cache miss", not so scary. By the way, after a second call,
+                // there is a probability of receiving PERSISTED_QUERY_NOT_FOUND error again, but it is so ghostly
+                // that this can and should be neglect so not to complicate the code without the need.
+                request.Query = savedQuery;
+                return await SendHttpRequestAsync<TResponse>(request, cancellationToken);
+            }
+            else if (response.Errors.Any(error => string.Equals(error.Message, "PersistedQueryNotSupported") ||
+                     response.Errors.Any(error => string.Equals(error.Message, "GraphQL query is missing.")))) // GraphQL.NET specific error message
```
Shane32, line 162: What if it's another GraphQL server which has a slightly different response?  This repo is not exclusively for GraphQL servers that either (1) support APQ or (2) are GraphQL.NET.  I'm not sure, but even GraphQL.NET might return different messages for older versions.

Shane32:
> please review, especially TODO note about extensions type

I use my own GraphQL client implementation, and it is based on generics; I wouldn't have a dictionary myself.  Although the serializer would handle a dictionary if it was passed it.  The same may be true here (?)

src/GraphQL.Client/GraphQLHttpClient.cs
sungam3r: I know. It's more like demo to discuss. My implementation has more conditions to consider - http status codes, graphql error codes in extensions, etc. - but I minimized them to minimum to make code clean. I'm open to suggestions.

sungam3r:
Problem with extensions is in design. What if someone set request.Extensions = new Person() before passing request into client? How to deal with arbitrary initialized object? How to **add** new data into it? Ping @rose-a , please comment.

rose-a:
@Shane32 how would you handle a generic extensions type when a user want's to use his own extension in combination with the integrated APQ feature? 

I think making `request.Extension` a `Dictionary<string, object?>` would be sensible in that it would provide a relatively safe way of providing integrated functionality like APQ and allowing the user to add other extensions without bothering too much with the integrated stuff (of course you can still provoke conflicts, but the chances are a lot smaller...). But there will propably be someone who's just sending a string via the extensions "object" right now who will complain anyway 😉 .

It also fits the [GraphQLOverHTTP Spec Draft](https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#request-parameters) where its specified as a map...

sungam3r:
Then I suggest to open another PR changing extensions property to `Dictionary<string, object?>`. OK?

Also note that the problem is not only between built-in functionality (like APQ or something else) and user extensions. User may use graphql client in such a way that different modules may want to set their own extensions and conflict on shared `object`. Dictionary (map) allows easily eliminate conflict.

src/GraphQL.Client/GraphQLHttpClient.cs
rose-a: Apollo uses a delegate to determine if it should disable persistent queries for the session... 

> `disable`: a function which takes an `ErrorResponse` (see below) and returns a boolean to disable any future persisted queries for that session. This defaults to disabling on `PersistedQueryNotSupported` or a 400 or 500 http error.

We'd have to handle the `GraphQLHttpRequestException` from `SendHttpRequestAsync` there, too.

src/GraphQL.Client/GraphQLHttpClientOptions.cs
sungam3r: Ok, I'll change to false.

Shane32:
> @Shane32 how would you handle a generic extensions type when a user want's to use his own extension in combination with the integrated APQ feature?

I don't use this library, but for my library, I'd serialize the user's anonymous object to a JsonElement and append the APQ data.

Let's ask ChatGPT for a sample:

## ChatGPT 4 request

Write a C# function with this signature:

Dictionary<string, object> AddAPQSupport(object extensions, int version, string sha256Hash)

The function should serialize 'extensions' to a JsonElement, expecting a map, and add two keys to the map, "version" and "sha256Hash".  If extensions is null it should create a new map

## ChatGPT 4 response

```cs
using System;
using System.Collections.Generic;
using System.Text.Json;

public class APQSupport
{
    public static Dictionary<string, object> AddAPQSupport(object extensions, int version, string sha256Hash)
    {
        Dictionary<string, object> resultMap;

        if (extensions == null)
        {
            resultMap = new Dictionary<string, object>();
        }
        else
        {
            try
            {
                string jsonString = JsonSerializer.Serialize(extensions);
                resultMap = JsonSerializer.Deserialize<Dictionary<string, object>>(jsonString);
            }
            catch (Exception)
            {
                throw new ArgumentException("Invalid 'extensions' input. It must be a serializable map.");
            }
        }

        resultMap["version"] = version;
        resultMap["sha256Hash"] = sha256Hash;

        return resultMap;
    }
}
```

> This C# function creates a new map if the extensions parameter is null or serializes the input to a JsonElement and converts it to a Dictionary. Then, it adds "version" and "sha256Hash" keys to the map. If the input is not a serializable map, it throws an ArgumentException.

Shane32:
Keep in mind that a dictionary works just as well to be fed into this as an anonymous type.  Now in my own library, I also support pre-encoded JSON strings as inputs to extensions/variables, not just anonymous types.  Just because we are having fun here, let's have ChatGPT add that feature, and also optimize it a bit:

# Request 2

change it so that if a string is passed in to extensions, it is assumed to be json and deserialized

# Request 3

utilize the SerializeToElement method so it does not need to be serialized to a string

# Final ChatGPT 4 output

```cs
using System;
using System.Collections.Generic;
using System.Text.Json;

public class APQSupport
{
    public static Dictionary<string, object> AddAPQSupport(object extensions, int version, string sha256Hash)
    {
        Dictionary<string, object> resultMap;

        if (extensions == null)
        {
            resultMap = new Dictionary<string, object>();
        }
        else
        {
            try
            {
                JsonElement jsonElement;

                if (extensions is string jsonString)
                {
                    jsonElement = JsonDocument.Parse(jsonString).RootElement;
                }
                else
                {
                    jsonElement = JsonSerializer.SerializeToElement(extensions);
                }

                resultMap = JsonSerializer.Deserialize<Dictionary<string, object>>(jsonElement.GetRawText());
            }
            catch (Exception)
            {
                throw new ArgumentException("Invalid 'extensions' input. It must be a serializable map or a valid JSON string.");
            }
        }

        resultMap["version"] = version;
        resultMap["sha256Hash"] = sha256Hash;

        return resultMap;
    }
}
```

Shane32:
It still doesn't deserialize directly from the JsonElement.  Just deleting `.GetRawText()` would give me the answer I was looking for.  That is a relatively new feature, not existing in .NET prior to version 6, so probably the ChatGPT AI doesn't know about it since it's knowledge cutoff is 2021 I think.

https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserialize?view=net-7.0#system-text-json-jsonserializer-deserialize-1(system-text-json-jsonelement-system-text-json-jsonserializeroptions)

Shane32:
@rose-a If I remember right, this library supports multiple serializers, so while this answer works for me, may not be the best answer here.  It's just an idea.

rose-a:
mh... while this solution would propably work its really hacky in my opinion and it results in a `Dictionary<string, object?>` anyway. 

So if someone puts something into the extensions object which is not serializable to a map it explodes at runtime instead of during or before compilation. (It propably would explode on the server side, anyway, at least with GraphQL.NET, as that uses an `Inputs` object which is a `ReadOnlyDictionary<string, object?>`)

src/GraphQL.Client/GraphQLHttpClient.cs
mikocot: the error handling is a bit of a pain, there doesn't seem to be any error code or structured response indicating missing or incorrect hash, I'm afraid we must just link currently known responses, for particular implementations to behaviors we want. Maybe at one point they will come up with a generic way to report such errors.

mikocot:
> Problem with extensions is in design. What if someone set request.Extensions = new Person() before passing request into client? How to deal with arbitrary initialized object? How to **add** new data into it? Ping @rose-a , please comment.

other option is to make it a base type (with only persisted query as property), that one may inherit from, and add additional properties. It's not as flexible as anonymous objecta as many currently probably use but a bit cleaner than making it a dictionary. In any case that seems to be a breaking change in the API

src/GraphQL.Client/GraphQLHttpClient.cs
```diff
@@ -111,6 +116,62 @@ public IObservable<GraphQLResponse<TResponse>> CreateSubscriptionStream<TRespons
...
+            useAPQ = true;
+            // TODO: I suggest to change request.Extension type to public Dictionary<string, object?>?
+            //request.Extensions ??= new();
+            //request.Extensions["persistedQuery"] = new Dictionary<string, object>
+            //{
+            //    ["version"] = APQ_SUPPORTED_VERSION,
+            //    ["sha256Hash"] = Hash.Compute(request.Query),
+            //};
+            request.Extensions ??= new
+            {
```
mikocot, line 139: doesn't it need those values to be wrapped within a 'PersistedQuery' property?

src/GraphQL.Client/Hash.cs
```diff
@@ -0,0 +1,44 @@
+using System.Buffers;
+using System.Diagnostics;
+using System.Security.Cryptography;
+using System.Text;
+
+namespace GraphQL.Client.Http;
+
+internal static class Hash
+{
+    private static SHA256? _sha256;
```
mikocot, line 10: sorry for my ignorance, but why using a shared static variable and deal with potential concurrency issue instead of just using one in local scope, not sure if that'd be a performance gain with many concurrent requests, and without many, well, it doesn't really matter.

src/GraphQL.Client/GraphQL.Client.csproj
```diff
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0;net461</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;net461;net5</TargetFrameworks>
```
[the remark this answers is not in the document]
Shane32: This just is used for feature targeting. So .NET 5 features only get used for .NET 5 and newer. It does not matter that it’s not supported.

src/GraphQL.Client/GraphQLHttpClient.cs
```diff
@@ -111,6 +116,62 @@ public IObservable<GraphQLResponse<TResponse>> CreateSubscriptionStream<TRespons
...
+                // that this can and should be neglect so not to complicate the code without the need.
+                request.Query = savedQuery;
+                return await SendHttpRequestAsync<TResponse>(request, cancellationToken);
+            }
+            else if (response.Errors.Any(error => string.Equals(error.Message, "PersistedQueryNotSupported") ||
+                     response.Errors.Any(error => string.Equals(error.Message, "GraphQL query is missing.")))) // GraphQL.NET specific error message
+            {
+                // GraphQL server either supports APQ of some other version, or does not support it at all.
+                // Send a request for the second time. This is better than returning an error. Let the client work with APQ disabled.
+                _useAPQ = false;
```
mikocot, line 166: it might be important to specify in the docs that graphql should be provisioned as singleton in such case, or this process would repeat for every request e.g. in asp case

src/GraphQL.Client/GraphQLHttpClient.cs
```diff
@@ -111,6 +116,62 @@ public IObservable<GraphQLResponse<TResponse>> CreateSubscriptionStream<TRespons
...
+        if (useAPQ && response.Errors?.Length > 0)
+        {
+            if (response.Errors.Any(error => string.Equals(error.Message, "PersistedQueryNotFound", StringComparison.CurrentCultureIgnoreCase)))
+            {
+                // GraphQL server supports APQ!
+
+                // Alas, for the first time we did not guess and in vain removed Query, so we return Query and
+                // send request again. This is one-time "cache miss", not so scary. By the way, after a second call,
+                // there is a probability of receiving PERSISTED_QUERY_NOT_FOUND error again, but it is so ghostly
+                // that this can and should be neglect so not to complicate the code without the need.
```
mikocot, line 157: there is a small chance this would lead to an infinite loop, if we imagine that service is misconfigured and e.g. has issues with caching (or doesnt add new values due to lack of memory?). We prob should give up after several tries? But I agree it adds a lot of complexity and might be not worth it

src/GraphQL.Client/GraphQL.Client.csproj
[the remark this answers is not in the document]
sungam3r: I see such a remark in different forms for the umpteenth time. net5 is a TFM describing specific API set. I need new APIs from net5, not from net7. What you wrote matters only for apps, not libraries.

src/GraphQL.Client/GraphQLHttpClient.cs
sungam3r: yep, forgot about 'PersistedQuery' when copy-pasted code. Operation name is out of scope here, APQ has nothing to do with operatio name.

src/GraphQL.Client/GraphQLHttpClientOptions.cs
sungam3r: done

src/GraphQL.Client/Hash.cs
sungam3r: It's a small optimization copied from GraphQL.NET. Actually, all code in that file comes from GraphQL.NET https://github.com/graphql-dotnet/graphql-dotnet/blob/master/src/GraphQL.MemoryCache/AutomaticPersistedQueries/AutomaticPersistedQueriesExecutionBase.cs#L57

src/GraphQL.Client/GraphQLHttpClient.cs
sungam3r: It's a general design question not coupled with APQ.
https://github.com/graphql-dotnet/graphql-client/issues/516
`GraphQLHttpClient` is designed to be singleton for endpoint. @rose-a would you like me to add such a note into README?

src/GraphQL.Client/GraphQLHttpClient.cs
sungam3r: I think it's better to remove that comment at all.

src/GraphQL.Client/GraphQLHttpClient.cs
rose-a: Yep that wouldn't hurt

src/GraphQL.Client/Hash.cs
```diff
@@ -0,0 +1,44 @@
...
+
+    internal static string Compute(string query)
+    {
+        var expected = Encoding.UTF8.GetByteCount(query);
```
Shane32, line 14: Might be better to use GetMaxByteCount here if you're trying to super-optimize this method.

sungam3r: I'm not trying to super-optimize, I just copied code from the main repo.
`query { ping }` yields 14 for `Encoding.UTF8.GetByteCount` and 45 for `Encoding.UTF8.GetMaxByteCount`. I'm fine either way but would like to have the same methods in both repos.

Shane32: Up to you.  Fine as-is

src/GraphQL.Client/GraphQLHttpResponse.cs
```diff
@@ -19,6 +19,13 @@ public GraphQLHttpResponse(GraphQLResponse<T> response, HttpResponseHeaders resp
     public HttpStatusCode StatusCode { get; set; }
 }
 
+public interface IGraphQLHttpResponse : IGraphQLResponse
```
sungam3r, line 22: This interface is needed to pass non-generic instance of response into `GraphQLHttpClientOptions.DisableAPQ`.

src/GraphQL.Client/GraphQLHttpClient.cs
sungam3r: > We'd have to handle the GraphQLHttpRequestException from SendHttpRequestAsync there, too

Still in TODO.

src/GraphQL.Client/GraphQLHttpResponse.cs
rose-a: I think you should put the content in there, too... then people can evaluate plaintext error messages...

**Always** having the raw content in there might be beneficial for general debugging, too, cause then people could cast their responses to `IGraphQLHttpResponse` and get the raw body of the response...

sungam3r: Then I would put entire `HttpResponseMessage` instead of pulling discrete properties from it. Also note that content may be already unaccessable since stream was read in `SendHttpRequestAsync`.

sungam3r: #559 , not a goal for this PR

rose-a (commented):
@sungam3r I picked up your work and made the following amendments:

- Refactor GraphQLHttpClient to allow usage of APQ via websocket connection, too
- Move hash/extension generation into *GraphQL.Primitives* and create a `GraphQLQuery` class which allows to declare a reusable query string with the hash computed only once during object creation.
- Add basic tests

@Shane32 @mikocot @sungam3r please review, I'd especially like your opinion on the current handling of the hash value in the [GraphQLRequest](https://github.com/graphql-dotnet/graphql-client/blob/apq/src/GraphQL.Primitives/GraphQLRequest.cs) class

The current implementation does not pre-compute the hash if it is created using the constructor accepting a string parameter, and it clears a hash taken from a `GraphQLQuery` constructor argument when setting the `Query` property.

This way there is no additional overhead to compute hashes in case APQ is not used, but if someone reuses the GraphQLRequest instance it would always be re-computed before sending. Alternatively we could compute the hash in the setter of the Query property, but that would then also happen if APQ is not used at all.

src/GraphQL.Primitives/GraphQLQuery.cs
```diff
@@ -1,15 +1,34 @@
...
 namespace GraphQL;
 
 /// <summary>
-/// Value record for a GraphQL query string
+/// Value object representing a GraphQL query string and storing the corresponding APQ hash. <br />
+/// Use this to hold query strings you want to use more than once.
 /// </summary>
-/// <param name="Text">the actual query string</param>
-public readonly record struct GraphQLQuery([StringSyntax("GraphQL")] string Text)
+public class GraphQLQuery : IEquatable<GraphQLQuery>
```
Shane32, line 8: Here's some different thoughts while looking at the changes here:
- Why moving from a struct to a class?
- Why not lazy load the hash?  Maybe more important to preload the hash if it was a struct, but we are using a class here.
- String comparisons will automatically short-circuit if they point to the same reference, so the benefit of comparing the Sha256Hash vs the Text is likely very minimal.

Not sure I have a concrete suggestion, but those are my thoughts.

src/GraphQL.Primitives/GraphQL.Primitives.csproj
```diff
@@ -6,4 +6,7 @@
     <TargetFrameworks>netstandard2.0;net6.0;net7.0;net8.0</TargetFrameworks>
   </PropertyGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="System.Memory" Version="4.5.5" />
```
Shane32, line 10: Where is this needed?  This reference is fine with me, but on the other hand, there's probably no reason to spend time optimizing older frameworks.  If users want an optimized implementation, they can upgrade to a modern version of .NET.  (Note: I am greatly in favor of supporting old frameworks via .NET Standard.)

src/GraphQL.Primitives/Hash.cs
```diff
@@ -0,0 +1,44 @@
...
+
+internal static class Hash
+{
+    private static SHA256? _sha256;
+
+    internal static string Compute(string query)
+    {
+        int expected = Encoding.UTF8.GetByteCount(query);
+        byte[]? inputBytes = ArrayPool<byte>.Shared.Rent(expected);
+        int written = Encoding.UTF8.GetBytes(query, 0, query.Length, inputBytes, 0);
```
Shane32, line 16: This could be rewritten to use Span / stackalloc also.

Shane32 (commented):
Here's a few comments

src/GraphQL.Primitives/GraphQLRequest.cs
```diff
@@ -59,16 +69,28 @@ public GraphQLRequest([StringSyntax("GraphQL")] string query, object? variables
...
+    public void GeneratePersistedQueryExtension()
+    {
+        if (Query is null)
+            throw new InvalidOperationException($"{nameof(Query)} is null");
+
+        Extensions ??= new();
+        Extensions[EXTENSIONS_PERSISTED_QUERY_KEY] = new Dictionary<string, object>
+        {
+            ["version"] = APQ_SUPPORTED_VERSION,
+            ["sha256Hash"] = _sha265Hash ?? Hash.Compute(Query),
```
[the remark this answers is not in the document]
Shane32: Note: no synchronization is needed here.  writing a reference is an atomic operation.  of course it could compute the hash on two different threads at once, but that's okay.

src/GraphQL.Primitives/GraphQLQuery.cs
rose-a: I had introduced the GraphQLQuery struct mainly to provide support the [StringSyntax("GraphQL")] attribute when declaring query strings as variables (so just a wrapper around the query string).

> Why moving from a struct to a class?

I went from struct to class because the reference type characteristics are (at least in my understanding) especially benefical for the APQ concept (i.e. declare ALL queries your application uses in one GraphQLQuery, send that once and then leverage APQ and operation name), thus reusing the same reference everywhere.

> Why not lazy load the hash?

I think it makes sense for the intended use of the GraphQLQuery class to compute the hash on object construction. It's propably something you rather want to happen during application startup than later when you are waiting for a query.
If you don't want to pre-compute the hash, just don't use the GraphQLQuery class and go with strings.

src/GraphQL.Primitives/GraphQL.Primitives.csproj
rose-a: The current implementation of the Hash class uses `System.Buffers` which gets pulled in by `System.Memory`. I think just referencing `System.Buffers` would be better here.

src/GraphQL.Primitives/Hash.cs
rose-a: Could you please elaborate? 😉

Shane32: I saw that stackalloc was used on lines 21-27.  Stackalloc could be used for the call to GetBytes as well.  This is faster than ArrayPool.

https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/operators/stackalloc

The easy recommended pattern is to write code like this:

```cs
int length = 1000;
Span<byte> buffer = length <= 1024 ? stackalloc byte[length] : new byte[length];
```

This assumes that typically the buffer length is relatively small, and in such cases, is ideal.  It also includes a fallback for large buffer requirements; but then you'd probably lose the ArrayPool (?).  And if it was known that the query would commonly be larger than you'd want to allocate on the stack, then maybe the cost of the heap allocation for the long queries might be worse than the benefits from stackalloc.  Or you'd have to write a separate if block for the ArrayPool path vs not.  I'm not really sure; maybe better off leaving it as-is.

Another possible optimization would be to guess at the length, rather than using GetByteCount.  You'd have to write a fallback routine in case the encoded string length was longer than the preallocated buffer though, and you might need to use the underlying Encoder rather than the static method.

src/GraphQL.Primitives/GraphQLQuery.cs
Shane32: Makes sense

src/GraphQL.Primitives/Hash.cs
rose-a: I think I'll leave it as it is. Using `Span<byte>` would require separate code for old frameworks again, and the best optimization is using the `GraphQLQuery` class to compute the hash only once... but anyway, thanks for the explanation!

Issue #554, opened 2023-04-24
Trying to parse non-json responses - 'U' is an invalid start of a value. LineNumber: 0


mikocot:
When using persisted queries the server is expected to return BadRequest whenever hash isn't correct or wasn't registered yet.
Such responses are part of the process and come as plain text (html/text). This means they obviously can't be deserialized to any response type using neither System.Text.Json nor Newtonsoft, not even to an object.

At the same time the default `IsValidResponseToDeserialize` seems to be perfectly fine with passing such responses to the serializer.
I can understand originally this behavior was introduced to allow for parsing validation errors (https://github.com/graphql-dotnet/graphql-client/pull/419), but I don't think it was meant to be as permissive (disregard content type).

A workaround is obviously to provide an alternative implementation, but I simply don't see a reasony why one would ever pass something that is not json-like to a json serializer, while the method is very precisely called: `IsValidResponseToDeserialize`. BadRequest html/text is neither valid nor possible to deserialize.

Current implementation:
```c#
    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r =>
        // Why not application/json? See https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#processing-the-response
        r.IsSuccessStatusCode || r.StatusCode == HttpStatusCode.BadRequest || r.Content.Headers.ContentType?.MediaType == "application/graphql+json";
```

sungam3r:
> Such responses are part of the process and come as plain text (html/text).

Could you please provide a link to specification describing that response format?

mikocot:
Well, even the docs linked in the code indicate that:
https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#processing-the-response

In our apollo server we get most of the errors as well formatter json and errors field, but there are some, e.g. related to incorrect query hash that come back as plain text. Possibly they are coming from some earlier layers?

sungam3r:
The docs you provided indicate that status code should be 400. I do not see requirement that response should be **plain text (html/text)**. On the contrary, in apollo docs I see 
![изображение](https://user-images.githubusercontent.com/21261007/234300016-013d2343-cce7-410b-8277-e4db90ee2cb6.png)

[...]

PR #557, opened 2023-04-26
Add API approval tests

sungam3r:

PR #560, opened 2023-04-27
Improve default IsValidResponseToDeserialize implementation

mikocot:
As mentioned in issue https://github.com/graphql-dotnet/graphql-client/issues/554 the default implementation of `IsValidResponseToDeserialize` currently does not verify if the response actually can be deserialized, regardless of its name.

As it's a logical OR it completely ignores the ContentType as long s the status matches Ok or BadRequest. At the same time it's possible that a server or any other component even before reaching the server returns one of those statuses in incorrect ContentType. Such behavior ends up with an ugly runtime exception, while it can be easily handled.

I also made this method more readable and debuggable. It's discussable if AcceptedResponseContentTypes  should be editable and exposed.

### DIFF
--- a/src/GraphQL.Client/GraphQLHttpClientOptions.cs
+++ b/src/GraphQL.Client/GraphQLHttpClientOptions.cs
@@ -61,9 +61,18 @@ public class GraphQLHttpClientOptions
     /// Note that compatible to the draft graphql-over-http spec GraphQL Server MAY return 4xx status codes (401/403, etc.)
     /// with well-formed GraphQL response containing errors collection.
     /// </summary>
-    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r =>
+    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = DefaultIsValidResponseToDeserialize;
+
+    internal static IReadOnlyCollection<string> AcceptedResponseContentTypes { get; } = new[] { "application/graphql+json", "application/json", "application/graphql-response+json" };
+
+    private static bool DefaultIsValidResponseToDeserialize(HttpResponseMessage r)
+    {
         // Why not application/json? See https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#processing-the-response
-        r.IsSuccessStatusCode || r.StatusCode == HttpStatusCode.BadRequest || r.Content.Headers.ContentType?.MediaType == "application/graphql+json";
+        if (r.Content.Headers.ContentType?.MediaType != null && !AcceptedResponseContentTypes.Contains(r.Content.Headers.ContentType.MediaType))
+            return false;
+
+        return r.IsSuccessStatusCode || r.StatusCode == HttpStatusCode.BadRequest;
+    }
 
     /// <summary>
     /// This callback is called after successfully establishing a websocket connection but before any regular request is made.
--- a/src/GraphQL.Client/GraphQLHttpClientOptions.cs
+++ b/src/GraphQL.Client/GraphQLHttpClientOptions.cs
@@ -63,7 +63,7 @@ public class GraphQLHttpClientOptions
     /// </summary>
     public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = DefaultIsValidResponseToDeserialize;
 
-    internal static IReadOnlyCollection<string> AcceptedResponseContentTypes { get; } = new[] { "application/graphql+json", "application/json", "application/graphql-response+json" };
+    private static IReadOnlyCollection<string> AcceptedResponseContentTypes { get; } = new[] { "application/graphql+json", "application/json", "application/graphql-response+json" };
 
     private static bool DefaultIsValidResponseToDeserialize(HttpResponseMessage r)
     {
--- a/src/GraphQL.Client/GraphQLHttpClientOptions.cs
+++ b/src/GraphQL.Client/GraphQLHttpClientOptions.cs
@@ -67,7 +67,6 @@ public class GraphQLHttpClientOptions
 
     private static bool DefaultIsValidResponseToDeserialize(HttpResponseMessage r)
     {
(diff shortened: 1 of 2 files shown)


src/GraphQL.Client/GraphQLHttpClientOptions.cs
```diff
@@ -61,9 +61,18 @@ public class GraphQLHttpClientOptions
...
     /// with well-formed GraphQL response containing errors collection.
     /// </summary>
-    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r =>
+    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = DefaultIsValidResponseToDeserialize;
+
+    internal static IReadOnlyCollection<string> AcceptedResponseContentTypes { get; } = new[] { "application/graphql+json", "application/json", "application/graphql-response+json" };
+
+    private static bool DefaultIsValidResponseToDeserialize(HttpResponseMessage r)
+    {
         // Why not application/json? See https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#processing-the-response
```
sungam3r, line 70:
```suggestion
```

src/GraphQL.Client/GraphQLHttpClientOptions.cs
```diff
@@ -61,9 +61,18 @@ public class GraphQLHttpClientOptions
     /// Note that compatible to the draft graphql-over-http spec GraphQL Server MAY return 4xx status codes (401/403, etc.)
     /// with well-formed GraphQL response containing errors collection.
     /// </summary>
-    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r =>
+    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = DefaultIsValidResponseToDeserialize;
+
+    internal static IReadOnlyCollection<string> AcceptedResponseContentTypes { get; } = new[] { "application/graphql+json", "application/json", "application/graphql-response+json" };
```
sungam3r, line 66: internal?

sungam3r:
>  It's discussable if AcceptedResponseContentTypes should be editable and exposed.

I'm fine to keep it `private` for now.

tests/GraphQL.Client.Serializer.Tests/DefaultValidationTest.cs
```diff
@@ -0,0 +1,30 @@
...
+using FluentAssertions;
+using GraphQL.Client.Http;
+
+using Xunit;
+
+namespace GraphQL.Client.Serializer.Tests;
+
+public class DefaultValidationTest
+{
+
```
sungam3r, line 12:
```suggestion
```

src/GraphQL.Client/GraphQLHttpClientOptions.cs
rose-a, line 66:
```suggestion
    private static IReadOnlyCollection<string> AcceptedResponseContentTypes { get; } = new[] { "application/graphql+json", "application/json", "application/graphql-response+json" };
```

yes, please change that to private

rose-a (changes requested):
Thanks for you contribution, please apply the suggested changes, then I'm fine to merge this...

mikocot:
> Thanks for you contribution, please apply the suggested changes, then I'm fine to merge this...

Thanks, I believe it's all corrected now.

sungam3r (approved):

src/GraphQL.Client/GraphQLHttpClientOptions.cs
```diff
@@ -61,9 +61,17 @@ public class GraphQLHttpClientOptions
     /// Note that compatible to the draft graphql-over-http spec GraphQL Server MAY return 4xx status codes (401/403, etc.)
     /// with well-formed GraphQL response containing errors collection.
     /// </summary>
-    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = r =>
-        // Why not application/json? See https://github.com/graphql/graphql-over-http/blob/main/spec/GraphQLOverHTTP.md#processing-the-response
-        r.IsSuccessStatusCode || r.StatusCode == HttpStatusCode.BadRequest || r.Content.Headers.ContentType?.MediaType == "application/graphql+json";
+    public Func<HttpResponseMessage, bool> IsValidResponseToDeserialize { get; set; } = DefaultIsValidResponseToDeserialize;
+
+    private static IReadOnlyCollection<string> AcceptedResponseContentTypes { get; } = new[] { "application/graphql+json", "application/json", "application/graphql-response+json" };
```
rose-a, line 66:
```suggestion
    private static readonly IReadOnlyCollection<string> AcceptedResponseContentTypes = new[] { "application/graphql+json", "application/json", "application/graphql-response+json" };
```

Lets reuse that collection... (please adapt the filed name to the configured rules after changing that)

rose-a (changes requested):

src/GraphQL.Client/GraphQLHttpClientOptions.cs
mikocot: yeah, good point, it was stuck in between of being a field and being actually configurable :P I'd make it a non static field of options possibly, but then we'd need to rewrite anther part of this class, not worth most likely

sungam3r (approved):

mikocot:
hm, something doesn't seem right in the workflow, it has timeout at the default GH time limit of 6h, while the previous run was ok

rose-a:
Yep, `GraphQL.Integration.Tests.dll` never finishes... Happening too in other PRs (like #557), but not locally...

mikocot:
hm, maybe a deadlock in xunit? I've seen with some asp.net core integration tests. It would only happen if several tests used the same classfixtures of custom WebApplicationFactory. In ended up to be bound to OTEL tracing and was enought to not configure it for testing.
However, the same might happen with other 'shared' components. According to xunit authors it's intentional... It might happen only on some machines depending on the amount of cores and runner setup. On the other hand when I run all the tests locally I get some failed ones, unrelated to my changes, so I might be having something missing in the environment, didn't have time to investigate.

mikocot:
can you restart the workflow? I dont seem to have rights

PR #572, opened 2023-06-02
Fix ImmutableConverter.cs

### LINKED ISSUE #571: ImmutableConverter bug when object has both constructor and properties
There's a bug in this code:
https://github.com/graphql-dotnet/graphql-client/blob/344a3e9f2ff2d110d331d377c25611e0306ed042/src/GraphQL.Client.Serializer.SystemTextJson/ImmutableConverter.cs#L36-L38

Expected:
result is true when all the properties have a corresponding parameter in the constructor.

Actual:
result is true when all of the constructor parameters have a corresponding property.

This results in the immutable converter being used for classes that have only part of their properties in the constructor. This is bad because then all the properties that are not set in the conbstructor are ignored.

JorisVanEijden:
Swap properties and parameters in applicability check of ImmutableConverter. This prevents properties from being lost when they are not all represented in the contructor as parameters.

Fixes #571

### DIFF
--- a/src/GraphQL.Client.Serializer.SystemTextJson/ImmutableConverter.cs
+++ b/src/GraphQL.Client.Serializer.SystemTextJson/ImmutableConverter.cs
@@ -19,31 +19,30 @@ public class ImmutableConverter : JsonConverter<object>
         if (nullableUnderlyingType != null && nullableUnderlyingType.IsValueType)
             return false;
 
-        bool result;
         var constructors = typeToConvert.GetConstructors(BindingFlags.Public | BindingFlags.Instance);
         if (constructors.Length != 1)
         {
-            result = false;
+            return false;
         }
-        else
+
+        var constructor = constructors[0];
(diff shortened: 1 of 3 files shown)


rose-a (changes requested):
Hi, thanks for your contribution! Please move the new unit test into [BaseSerializerTest](https://github.com/graphql-dotnet/graphql-client/blob/master/tests/GraphQL.Client.Serializer.Tests/BaseSerializerTest.cs), we should make sure this behavior is consistent across all serializers...

JorisVanEijden:
> Hi, thanks for your contribution! Please move the new unit test into [BaseSerializerTest](https://github.com/graphql-dotnet/graphql-client/blob/master/tests/GraphQL.Client.Serializer.Tests/BaseSerializerTest.cs), we should make sure this behavior is consistent across all serializers...

Done.

rose-a:
👍 I'll try to review it this week...

rose-a (approved):

Issue #630, opened 2024-04-03
Please consider adding [StringSyntax("GraphQL")] attributes


lol768:
.NET allows you to use the StringSyntaxAttribute to mark up properties that are being used to hold text that an IDE should consider an "injected language".

E.g. `GraphQLRequest.Query` could be marked as a GraphQL fragment, and tools such as [this VSCode plugin](https://marketplace.visualstudio.com/items?itemName=codearchitects-research.GraphQLTools) and other IDEs (e.g. JetBrains Rider) would be able to syntax highlight queries automatically.

rose-a:
Hi, good proposal, hadn't noticed the existence of the StringSyntaxAttribute yet.

Requires a multiframework build though, as its only available from .NET 7 onwards and the library is currently built for .NET Standard 2.0...

lol768:
That's a good point, my PR doesn't properly deal with that at the moment.

I've got a stub implementation of the attribute to enable it to build for .NET Standard 2.0 (or e.g. net6) but there's probably a bit of additional work to do with the build

lol768:
> Requires a multiframework build though, as its only available from .NET 7 onwards and the library is currently built for .NET Standard 2.0...

Have added multiframework `GraphQL.Primitives` build to the open PR.

rose-a:
added in v6.0.4

PR #631, opened 2024-04-03
Address #630

lol768:
See #630

### DIFF
--- a/src/GraphQL.Client.Abstractions/GraphQL.Client.Abstractions.csproj
+++ b/src/GraphQL.Client.Abstractions/GraphQL.Client.Abstractions.csproj
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
     <Description>Abstractions for GraphQL.Client</Description>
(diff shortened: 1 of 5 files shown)


lol768:
This works when I publish the package locally, now, and reference it in a `net8.0` project:

image


It still continues to work in `netstandard2.0` projects, but the IDE understandably doesn't know what the stub attribute means/is.

rose-a:
Hi, thanks for the contribution... please allow me some time for review, I'll hopefully manage to have a look at it within the next few days...

rose-a (changes requested):
THe attribute should be added to the extension methods directly accepting a query string, too.

https://github.com/graphql-dotnet/graphql-client/blob/master/src/GraphQL.Client.Abstractions/GraphQLClientExtensions.cs

rose-a (approved):

lol768:
Thanks for reviewing @rose-a 🙂

PR #633, opened 2024-04-15
Removed reference to GraphQL.Client.Serializer.Newtonsoft in GraphQL.Client.LocalExecution

### LINKED ISSUE #632: Unused project reference to Newtonsoft.Json in GraphQL.Client.LocalExecution
Was trying to remove `Newtonsoft.Json` as a transitive dependency and found that this project has a reference to it, but it is not used anywhere, removed it and ran some tests and it seemed to work fine, not sure if there is a reason for it.

https://github.com/graphql-dotnet/graphql-client/blob/3e062755e952bbdc20b3158e17ecb131214c6103/src/GraphQL.Client.LocalExecution/GraphQL.Client.LocalExecution.csproj#L15

Meantub:
Fixes #632

Noticed that this might be something unused, let me know if it's there for some purpose


rose-a:
Hi, thanks for your contribution. This was most likely a leftover from when Newtonsoft was hardcoded inside the local execution client.

PR #673, opened 2026-01-02
Add missing cancellation tokens to async stream operations

extra2:
Fixed missing CancellationToken parameters in GraphQLHttpClient.cs:
- Pass cancellationToken to ReadAsStreamAsync() on line 180
- Pass cancellationToken to ReadToEndAsync() on line 193

This ensures proper cancellation support throughout the HTTP request pipeline, allowing operations to be cancelled promptly when requested.


JR-Morgan:
These overloads are only available in .NET core right?
Have you tried building the .NET Standard 2.0 target?

I suspect something like this would be needed.
```csharp
#if NET8_0_OR_GREATER
//ReadAsStreamAsync(cancellationToken);
#else
//ReadAsStreamAsync();
#endif
````

